PR #338, opened 2018-12-07
[5.3] environment resource

a-palchikov:
Bump satellite to support serf client reconnect.
Also fixes a minor annoyance with coredns [log spam](https://github.com/kubernetes/kubernetes/issues/17162).
Complements https://github.com/gravitational/gravity/pull/135.

Requires https://github.com/gravitational/satellite/pull/89.


lib/box/cfg.go
```diff
@@ -383,6 +400,44 @@ func (d *Devices) String() string {
...
+func ReadEnvironment(path string) (vars EnvVars, err error) {
+	f, err := os.Open(path)
+	if err != nil {
+		return nil, trace.ConvertSystemError(err)
+	}
+	defer f.Close()
+	return ReadEnvironmentFromReader(f)
+}
+
+func ReadEnvironmentFromReader(r io.Reader) (vars EnvVars, err error) {
```
r0mant: Godoc.

lib/box/cfg.go
```diff
@@ -178,6 +182,19 @@ func (vars *EnvVars) Environ() (environ []string) {
 	return environ
 }
 
+// WriteTo writes the environment variables to the specified writer w.
+// WriteTo implements io.WriterTo
+func (vars EnvVars) WriteTo(w io.Writer) (n int64, err error) {
+	cw := &countingWriter{}
+	for _, v := range vars {
+		// quote value as it may contain spaces
+		if _, err := fmt.Fprintf(io.MultiWriter(w, cw), "%v=%q\n", v.Name, v.Val); err != nil {
```
r0mant: Doesn't `fmt.Fprintf` return number of bytes written? Why is a separate "counting writer" needed?

lib/utils/utils.go
```diff
@@ -56,29 +59,136 @@ func DropInDir(unit string) string {
...
+
+	cleanup := func() error {
+		err := os.Remove(tmp.Name())
+		if err != nil {
+			log.WithFields(log.Fields{
+				"error": err,
+				"file":  tmp.Name(),
+			}).Warn("Failed to remove.")
+		}
+		return trace.ConvertSystemError(err)
```
r0mant: The cleanup's return argument is not actually used anywhere in this function so it can very well be made niladic instead.

lib/utils/utils.go
```diff
@@ -56,29 +59,136 @@ func DropInDir(unit string) string {
...
+	}
+	err = os.Rename(tmp.Name(), dst)
+	if err != nil {
+		cleanup()
+		return trace.ConvertSystemError(err)
+	}
+	return nil
+}
+
+func ConvertError(err error, format string, args ...interface{}) error {
```
r0mant: Godoc.

tool/planet/environment.go
```diff
@@ -0,0 +1,198 @@
...
+
+func updateEnvironment(ctx context.Context, kvs map[string]string) error {
+	log.WithField("kvs", kvs).Info("Update environment.")
+	env, err := box.ReadEnvironment(ContainerEnvironmentFile)
+	if err != nil {
+		return trace.Wrap(err, "failed to read cluster environment file")
+	}
+
+	for k, v := range kvs {
+		env.Upsert(k, v)
```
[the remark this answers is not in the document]
r0mant: I think this is a good point. How do you delete a variable that you, for example, previously set?

tool/planet/environment.go
```diff
@@ -0,0 +1,100 @@
...
+	if err != nil {
+		return trace.Wrap(err)
+	}
+
+	isLeader := env.Get(EnvPublicIP) == leaderAddr
+	services := append(commonServices, kubeServices...)
+	if isLeader {
+		services = append(services, kubeMasterServices...)
+	}
+	return restartServices(context.Background(), services)
```
r0mant: Have you considered taking even more granular approach i.e. have separate phases for restart of each service? Then if a certain service failed to restart, it will be easy to see what failed and rollback. And this function would only update the container-environment file. With this "batch restart" approach, wouldn't it be harder to rollback should something goes wrong?

tool/planet/environment.go
```diff
@@ -0,0 +1,100 @@
...
+	env, err := box.ReadEnvironment(ContainerEnvironmentFile)
+	if err != nil {
+		return trace.Wrap(err, "failed to read cluster environment file")
+	}
+
+	for k, v := range kvs {
+		env.Upsert(k, v)
+	}
+
+	err = utils.SafeWriteFile(ContainerEnvironmentFile, env, constants.SharedReadMask)
```
r0mant: Do the changes to the container-environment file persist between planet restarts / reboots / upgades/rollbacks? I think they do but I always forget...

tool/planet/environment.go
a-palchikov: I have and I don't think it would work. The first issue is that I cannot seem to figure out how to handle rollbacks in proper order given separate steps for services:
i.e. normal order:
  1. update container environment file
  1. restart service 1
  1. ...
  1. restart service n

For rollback, I'd still need step 1 before the rest - it does not work like that now as rollback steps are executed last to first. Besides, since requirements are merely recommendations and are only validated for forward execution, the rollbacks do not take them into account. Not sure how much time it would cost to change that as well w/o breaking the rest of operations.

tool/planet/environment.go
a-palchikov: That's another good point - they don't (I think you know they don't) - each time planet starts, it would recreate the environment file.
I see a couple of options:
  - add back direct ConfigMap access to planet so it could pull actual / previous values and update the file accordingly
  - update the planet's configuration package on each envar update operation. This will keep planet unaware of the environment variables setup in kubernetes (not sure to advantage though).

PR #352, opened 2018-12-19
Add node labels to kubelet.

r0mant:


tool/planet/cfg.go
```diff
@@ -113,6 +113,8 @@ type Config struct {
 	DNS DNS
 	// Taints is a list of kubernetes taints to apply to the object
 	Taints []string
+	// Labels is Kubernetes node labels
```
a-palchikov: `NodeLabels`

PR #371, opened 2019-01-17
Kevin/master/3898 service route

knisbet:
Updates: https://github.com/gravitational/planet/pull/new/kevin/master/3898-service-route

I'm not sure if there is a better way to tackle this. Create a dummy interface, and use it to null route the service subnet IP range. I tried to route towards the existing flannel interface, but this appears to block the traffic before netfilter applies the NAT for some reason.

Risk - On some customer networks the service subnet might overlap with customer network ranges. Adding this route, should act as a blackhole for the entire service subnet, which means the entire range will no long route within the customers network. This should be acceptable, as this is always a risk since services will be randomly assigned in this range anyways.

Thoughts?

### DIFF
--- a/build.assets/makefiles/base/network/wait-for-flannel.sh
+++ b/build.assets/makefiles/base/network/wait-for-flannel.sh
@@ -10,7 +10,7 @@ done;
 # The node may not have a default gateway or routes that cover the service subnet
 # This prevents "no route to host" errors when trying to reach the service subnet
 # by creating a dummy interface, and routing the service subnet to this interface
-# we can guarentee that the route exists, and can be NAT to the correct destination
+# we can guarantee that the route exists, and can be NAT to the correct destination
 ip link add flannel.null type dummy
 ip link set flannel.null up
 ip route add ${KUBE_SERVICE_SUBNET} dev flannel.null
--- a/build.assets/makefiles/base/network/wait-for-flannel.sh
+++ b/build.assets/makefiles/base/network/wait-for-flannel.sh
@@ -1,0 +1,17 @@
+#!/bin/bash
+source /etc/container-environment
+
+until [ -e /run/flannel/subnet.env ]; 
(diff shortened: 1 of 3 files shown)


build.assets/makefiles/base/network/wait-for-flannel.sh
```diff
@@ -0,0 +1,17 @@
...
+until [ -e /run/flannel/subnet.env ]; 
+  do echo \"waiting for flannel to start.\"; 
+  sleep 3; 
+done; 
+
+# https://github.com/gravitational/gravity.e/issues/3898
+# The node may not have a default gateway or routes that cover the service subnet
+# This prevents "no route to host" errors when trying to reach the service subnet
+# by creating a dummy interface, and routing the service subnet to this interface
+# we can guarentee that the route exists, and can be NAT to the correct destination
```
klizhentas: guarantee

klizhentas:
the range will not be routable within the context of one machine, not the whole network though, so it should be fine

build.assets/makefiles/base/network/wait-for-flannel.sh
```diff
@@ -0,0 +1,17 @@
...
+  do echo \"waiting for flannel to start.\"; 
+  sleep 3; 
+done; 
+
+# https://github.com/gravitational/gravity.e/issues/3898
+# The node may not have a default gateway or routes that cover the service subnet
+# This prevents "no route to host" errors when trying to reach the service subnet
+# by creating a dummy interface, and routing the service subnet to this interface
+# we can guarantee that the route exists, and can be NAT to the correct destination
+ip link add flannel.null type dummy
```
a-palchikov: Is `dummy` module guaranteed to be available or should it be loaded if not?

knisbet: I've never seen it not be loaded and it worked on my systems. To be sure I'll add it to the gravity checkers.

PR #377, opened 2019-01-23
Support for metrics-server

UnderGreen:


build.assets/docker/os.dockerfile
```diff
@@ -9,32 +9,23 @@ ADD os-rootfs/ /
...
 	gnupg2 \
 	dirmngr \
 	; \
 	rm -rf /var/lib/apt/lists/*; \
 	fi
 
-# dockerproject debian repo key
-RUN apt-get update && apt-get -q -y install apt-transport-https
-
-RUN (echo 'deb http://httpredir.debian.org/debian/ stretch contrib non-free' >> /etc/apt/sources.list && \
```
UnderGreen: I don't know why we add those sources here, because of default sources.list contains all that we need. I just updated that list with adding `non-free contrib` in lines below.

PR #382, opened 2019-01-31
Kevin/master/3922 cloud config configmap

knisbet:
Updates: https://github.com/gravitational/gravity.e/issues/3922

Allows setting /etc/cloud-config.conf from a kubernetes configmap. I don't think it's necessary to do a rolling restart, as it only restarts the controller-manager, and not any other daemon's with the current implementation.


klizhentas:
Before merging this, I would like to understand:

* Could and should it be implemented as `gravity resource` following our same path as `Environment` resource. 
* How does it collaborate with `gravity plan`

knisbet:
Abandoning as per the discussion here: https://github.com/gravitational/gravity.e/issues/3922

PR #390, opened 2019-02-19
Kevin/experimental/cgroups3

knisbet:
This PR fixes a number of issues I've been having with the planet container.

Most notable:
- switch to using systemd to create container cgroup
    - This causes the container to run in a separate scope that has delegate=true (/system.slice/planet-<uuid>.scope) which should hopefully prevent a possible conflict in having multiple systemd instances running and modifying the same cgroup hierarchy.
- Pass the container id to planet systemd. Not very important now, but if we setup machine integration, this should eventually allow journalctl to reach the planet journal without having to enter the container.
- Set Init/CWD on creation. This seems to be important on updating to runc rc6, where I've been having issues with the planet init process not running in the correct namespaces.
- Use CGROUP namespace. The other fixes cause systemd to correctly realize it is running within a container, however, systemd when starting remounts the /sys/fs/cgroup filesystem hierarchy to only contain its own portion of the hierarchy. This appears to greatly confuse kubelet, since when reading cgroup paths of processes, they do not match the mounted hierarchy. So this places planet in it's own cgroup namespace, so that all planet processes have a consistent view of the cgroup hierarchy.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -558,7 +558,7 @@
   revision = "7d4729fb36185a7c1719923406c9d40e54fb93c7"
 
 [[projects]]
-  digest = "1:c8c1cb86074527519ceb2290cff4a78545cfa71844acaea9e670a7dc1354a01c"
+  digest = "1:3a7b076f85f7786b280e7b06017cfa7f9379dfde983ac5280eb471d0f8a14baa"
   name = "github.com/opencontainers/runc"
   packages = [
     "libcontainer",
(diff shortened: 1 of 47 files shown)


a-palchikov:
About the potential of using the host's journalctl to access the container journals - are you certain this works even with incompatible systemd builds? I know for fact that the `systemd` we use(d) in `debian` was compiled with compression support and that was the reason `journalctl` from host would fail to decode container journal files.

knisbet:
No I don't know if this will work consistently. It's a systemd feature for systemd's version of containers, so I would expect it to work between versions, but I have no idea how careful the systemd project is with cross version or backwards compatibility.

PR #392, opened 2019-02-27
Test upstream nameservers for reachability

knisbet:
Note: Not to be merged until after 5.5 maintenance branches are cut

This adds to planet, testing reachability to upstream name servers.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -334,7 +334,7 @@
   version = "v1.0.0"
 
 [[projects]]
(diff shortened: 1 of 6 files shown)


knisbet:
Sorry, need to abandon this one, the golang resolver doesn't allow queries for root nameservers :(

https://github.com/golang/go/blob/2012227b01020eb505cf1dbe719b1fa74ed8c5f4/src/net/dnsclient.go#L60-L115

I'll need to look into whether we can use an alternate resolver.

PR #397, opened 2019-03-02
cgroup config to prevent resource starvation

knisbet:
Requires: #390 

This implements the planned cgroup changes to try and address resource starvation problems on busy gravity clusters. 

Waiting for 5.5 maintenance branches.


tool/planet/cgroup.go
```diff
@@ -0,0 +1,315 @@
...
+type CgroupConfig struct {
+	// Enabled indicates whether the planet container should apply the cgroup configuration.
+	Enabled bool
+	// Auto indicates whether planet is allowed to regenerate the Cgroup configuration.
+	// If a new version of planet embeds new rules, this allows planet to update the configuration file. If the
+	// config is externally managed or updated by a user, set this to false so planet doesn't overwrite the settings.
+	Auto bool
+
+	// KubeReservedCPU is the amount of CPU to reserve within kubernetes for kubelet + docker (in millicores)
+	KubeReservedCPU int
```
a-palchikov: nit: `KubeReservedCPUMillicores` would make it self-documenting.

tool/planet/cgroup.go
```diff
@@ -0,0 +1,315 @@
...
+
+	// Cgroups is a list of cgroup configurations to apply within the planet container
+	Cgroups []CgroupEntry
+}
+
+type CgroupEntry struct {
+	specs.LinuxResources
+
+	// Path is the cgroup hierarchy path this setting applies to
+	Path []string `json:"path"`
```
a-palchikov: Why does this need a json tag and all other fields don't? Also it would be helpful to document why this is a string slice.

tool/planet/cgroup.go
```diff
@@ -0,0 +1,315 @@
...
+	err = writeCgroupConfig(configPath, config)
+	if err != nil {
+		return trace.Wrap(err)
+	}
+	var errors []error
+
+	// try and set the cgroups
+	for _, entry := range config.Cgroups {
+		if len(entry.Path) == 0 {
+			return trace.BadParameter("cgroup spec with no path set: %v", entry)
```
a-palchikov: Wouldn't this check defeat the purpose after the configuration has already been written to disk?

tool/planet/cgroup.go
```diff
@@ -0,0 +1,315 @@
...
+}
+
+func newProperty(name string, units interface{}) systemdDbus.Property {
+	return systemdDbus.Property{
+		Name:  name,
+		Value: dbus.MakeVariant(units),
+	}
+}
+
+func systemdPath(s ...string) cgroups.Path {
```
a-palchikov: Does not seem to be used for anything.

tool/planet/cgroup_test.go
```diff
@@ -0,0 +1,372 @@
...
+
+import (
+	"testing"
+
+	specs "github.com/opencontainers/runtime-spec/specs-go"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestDefaultCgroupConfig(t *testing.T) {
+	//swappiness := uint64(20)
```
a-palchikov: nit: remove commented out code

tool/planet/cgroup.go
knisbet: I don't think so, the configuration written to disk, is just so that it's possible for a user or external tool to edit the configuration for the next start of the node. As long as Enabled and Auto are true, the write to disk only really let's you go in and see the config that is being used. However, if it's user specified configuration, it may still have errors.

lib/box/enter.go
```diff
@@ -151,3 +159,54 @@ func StartProcessStdout(c libcontainer.Container, cfg ProcessConfig) error {
...
+
+	if !strings.HasPrefix(cgroupPath, "/sys/fs/cgroup") {
+		return trace.BadParameter("Cgroup path not mounted to /sys/fs/cgroup: ", cgroupPath)
+	}
+
+	// Example cgroup path: /sys/fs/cgroup/cpu,cpuacct/system.slice/-planet-cee2b8a0-c470-44a6-b7cc-1eefbc1cc88c.scope
+	// we want to split off the /sys/fs/cgroup/cpu,cpuacct/ part, so we have just the cgroup structure
+	// (system.slice/-planet-cee2b8a0-c470-44a6-b7cc-1eefbc1cc88c.scope)
+	dirs := strings.Split(cgroupPath, "/")
+	userPath := path.Join("/", path.Join(dirs[5:]...), "user")
```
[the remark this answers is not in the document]
knisbet: fixed

PR #398, opened 2019-03-05
Use separate kubeconfig for host kubectl.

r0mant:
When generating `kubectl.kubeconfig` planet uses secrets from `/var/lib/gravity` which does not work when kubectl is invoked from host and custom state directory was used for gravity data.

This PR addresses this issue by generating two configs: one will be used when kubectl is invoked from host and uses correct host state dir, another one - when kubectl is invoked from inside planet.

### DIFF
--- a/tool/planet/start.go
+++ b/tool/planet/start.go
@@ -546,18 +546,27 @@ func applyConfigOverrides(config *kubeletconfig.KubeletConfiguration) error {
 
 // addKubeConfig writes a kubectl config file
 func addKubeConfig(config *Config) error {
-	kubeConfig, err := NewKubeConfig(config.APIServerIP())
-	if err != nil {
-		return trace.Wrap(err)
+	// Generate two kubectl configuration files: one will be used by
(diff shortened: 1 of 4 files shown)


tool/planet/start.go
```diff
@@ -546,18 +546,27 @@ func applyConfigOverrides(config *kubeletconfig.KubeletConfiguration) error {
...
+		kubeConfig, err := NewKubeConfig(config.APIServerIP(), stateDir)
+		if err != nil {
+			return trace.Wrap(err)
+		}
+		path := filepath.Join(config.Rootfs, configPath)
+		err = os.MkdirAll(filepath.Dir(path), constants.SharedDirMask)
+		if err != nil {
+			return trace.Wrap(err)
+		}
+		err = ioutil.WriteFile(path, kubeConfig, constants.SharedReadMask)
```
a-palchikov: nit: we should stick to `utils.SafeWriteFile` unless unsuitable.

PR #405, opened 2019-03-28
bump CNI and kubernetes versions

knisbet:


build.assets/makefiles/base/network/network.mk
```diff
@@ -15,7 +15,7 @@ all: $(BINARIES) network.mk
 
 # Setup CNI and include flannel as a plugin
 	mkdir -p $(ROOTFS)/etc/cni/net.d/ $(ROOTFS)/opt/cni/bin
-	curl -L --retry 5 https://github.com/containernetworking/plugins/releases/download/v0.7.1/cni-plugins-amd64-v0.7.1.tgz \
+	curl -L --retry 5 https://github.com/containernetworking/plugins/releases/download/v0.7.5/cni-plugins-amd64-v0.7.5.tgz \
```
a-palchikov: nit: feels like this version can be moved to the version block in the Makefile

knisbet: Yea, I agree, just didn't want to mess with that on this particular patch.

PR #408, opened 2019-03-29
build registry off fork with hardened ciphers

knisbet:
Temporarily build registry from fork, which has been hardcoded to use modern TLS and ciphers.

Requires Backport

### DIFF
--- a/build.assets/makefiles/base/docker/registry.mk
+++ b/build.assets/makefiles/base/docker/registry.mk
@@ -1,7 +1,7 @@
 .PHONY: all
 
 REPODIR=$(GOPATH)/src/github.com/docker/
-VER=v2.6.0
+VER=v2.7.1-gravitational
 
 # VERSION_PACKAGE defines contents of a `version.go` file which is part of docker
 # registry source code distribution responsible for defining a registry's Version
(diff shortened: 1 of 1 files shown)


build.assets/makefiles/base/docker/registry.mk
```diff
@@ -1,7 +1,7 @@
 .PHONY: all
 
 REPODIR=$(GOPATH)/src/github.com/docker/
-VER=v2.6.0
+VER=v2.7.1-gravitational
```
r0mant: Is this version backwards-compatible with 2.6? IIRC last time we upgraded the registry minor version (2.4 -> 2.6 I think) we had to upgrade vendored library too b/c tele build was producing incompatible images. Am I correct @a-palchikov?

build.assets/makefiles/base/docker/registry.mk
```diff
@@ -30,7 +30,8 @@ all: $(BINARIES) install
 $(BINARIES):
 	@echo "\n---> Building docker registry:\n"
 	mkdir -p $(REPODIR)
-	cd $(REPODIR) && git clone https://github.com/docker/distribution -b $(VER) --depth 1
+#	cd $(REPODIR) && git clone https://github.com/docker/distribution -b $(VER) --depth 1
+	cd $(REPODIR) && git clone https://github.com/gravitational/distribution -b $(VER) --depth 1
```
r0mant: Add a comment about why we're building off fork and a TODO item to fix it?

build.assets/makefiles/base/docker/registry.mk
knisbet: According to the 2.7.1 release notes, they fixed an issue when going to 2.6.0 -> 2.7.1, but besides that I don't know if there might be more compatibility issues lurking.

PR #415, opened 2019-04-18
Fix a typo in kube-apiserver.service

itay:
When the option to set the `--service-node-port-range`, there was a typo in the environment variable the flags were appended to. This fixes it.


itay:
I was trying to figure out why setting `serviceNodePortRange: "3000-32767"` in `cluster-config.yaml` wasn't being picked up, and I'm pretty sure this is it.

It would be great if this could be backported to 5.4.x and 5.5.x as well, so that people who don't want to pick up the latest upgrade to Docker 18.09 can use it too.

a-palchikov:
Thanks for your contribution! I will gladly port the change to other versions.

itay:
Thanks @a-palchikov ! I'll be honest, I actually built a custom image with this but I still couldn't get this to work.

Have you gotten it to work with the `serviceNodePortRange: "3000-32767"` option in `cluster-config.yaml`?

PR #419, opened 2019-04-22
detect if kernel has cgroup ns support

knisbet:
Only enable resource starvation prevention when running on kernels with cgroup namespace support


r0mant:
Merged as a part of https://github.com/gravitational/planet/pull/414.

PR #424, opened 2019-05-21
http no_proxy

knisbet:
Planet equivelant change of https://github.com/gravitational/gravity/pull/416. Appends to customer provided no_proxy settings .local and 0.0.0.0/0 so that connections within the cluster do not attempt to invoke the proxy. Similar to the Gravity PR, 0.0.0.0/0 is overlay broad, but because the cluster configuration changes over time, it's difficult to predict all IP addresses we need to set within no_proxy. 

The other thing I had to do, is separate http_proxy related config into it's own environment file (set to /etc/proxy-environment). This is because at-least etcd, but likely other processes don't support cidr notation in the no_proxy environment due to using an outdated version of golang. As such, I also only pass the proxy settings to the dockerd process and shell session, which should be the only planet based processes that require outbound connectivity. If you can think of additional services let me know.

Updates https://github.com/gravitational/gravity.e/issues/3998

### DIFF
--- a/lib/box/cfg.go
+++ b/lib/box/cfg.go
@@ -135,6 +135,17 @@ func (vars *EnvVars) Get(v string) string {
 	return ""
 }
 
+// Delete removes the environment variable named v from the list and returns the value
+func (vars *EnvVars) Delete(v string) string {
+	for i, p := range *vars {
+		if p.Name == v {
+			*vars = append((*vars)[:i], (*vars)[i+1:]...)
+			return p.Val
+		}
+	}
+	return ""
+}
+
 // Append adds a new environment variable given with k, v and delimiter delim
 func (vars *EnvVars) Append(k, v, delim string) {
 	if existing := vars.Get(k); existing != "" {
--- a/lib/box/cfg.go
+++ b/lib/box/cfg.go
@@ -135,7 +135,7 @@ func (vars *EnvVars) Get(v string) string {
 	return ""
 }
 
-// Delete removes the environment variable named v from the list and returns the value
+// Delete removes the environment variable named v from the list and returns its value
 func (vars *EnvVars) Delete(v string) string {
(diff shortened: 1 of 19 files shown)


lib/box/cfg.go
```diff
@@ -135,6 +135,17 @@ func (vars *EnvVars) Get(v string) string {
 	return ""
 }
 
+// Delete removes the environment variable named v from the list and returns the value
```
r0mant: Nit: I'd say "returns **its** value".

tool/planet/start.go
```diff
@@ -843,6 +891,10 @@ const (
 	DockerWorkDir            = "/ext/docker"
 	RegistryWorkDir          = "/ext/registry"
 	ContainerEnvironmentFile = "/etc/container-environment"
+	// ProxyEnvironmentFile is an environment file with outbound proxy options
+	// Note: these settings are separates from container-environent because not all processes should load the proxy
```
r0mant: separates -> separate

tool/planet/start.go
```diff
@@ -349,6 +356,47 @@ func addCloudOptions(c *Config) error {
...
+
+	// Move proxy settings from regular Environment to Proxy environment settings so not all planet processes
+	// try and use the http_proxy.
+	for _, k := range proxy {
+		c.ProxyEnv.Upsert(k, c.Env.Delete(k))
+	}
+
+	// If we're unable to locate a NO_PROXY config, create default settings
+	if !found {
+		c.ProxyEnv.Upsert("NO_PROXY", strings.Join([]string{"0.0.0.0/0", ".local"}, ","))
```
r0mant: Maybe move `[]string{"0.0.0.0/0", ".local"}` into a some kind of constant (even if in this module) as it's being used in a few places?

PR #425, opened 2019-06-03
Allow quoted values for environment

a-palchikov:
Allow environment specification of the form `name=value,name2=value2` where `value` can itself contain a comma (or be basically free-form).

Updates https://github.com/gravitational/gravity.e/issues/4095.


a-palchikov:
Yes, gravity will take care of quoting if a value contains commas.

PR #428, opened 2019-06-04
http no_proxy (#424) backport 5.5.x

knisbet:


tool/planet/start.go
```diff
@@ -349,6 +356,49 @@ func addCloudOptions(c *Config) error {
...
+	}
+
+	noProxy := map[string]string{
+		"NO_PROXY": c.Env.Delete("NO_PROXY"),
+		"no_proxy": c.Env.Delete("no_proxy"),
+	}
+
+	staticNoProxyRules := []string{"0.0.0.0/0", ".local"}
+
+	found := false
```
a-palchikov: Like in the previous PR, I'd suggest to reorder the statements to get rid of the flag:
```go
func configureProxy(c *Config) {
	// Attempt to find any current proxy settings in the variable names golang supports
	// https://github.com/golang/net/blob/c21de06aaf072cea07f3a65d6970e5c7d8b6cd6d/http/httpproxy/proxy.go#L91-L98
	proxy := []string{
		"HTTP_PROXY",
		"http_proxy",
		"HTTPS_PROXY",
		"https_proxy",
	}

	noProxy := map[string]string{
		"NO_PROXY": c.Env.Delete("NO_PROXY"),
		"no_proxy": c.Env.Delete("no_proxy"),
	}

	staticNoProxyRules := []string{"0.0.0.0/0", ".local"}

	// Move proxy settings from regular Environment to Proxy environment settings so not all planet processes
	// try and use the http_proxy.
	moveEntries(&c.ProxyEnv, &c.Env, proxy)

	for k, v := range noProxy {
		if len(v) == 0 {
			continue
		}
		// TODO(knisbet) we should see if there is a way to not NO_PROXY every IP address
		// but it's difficult because we need to know each of the nodes IP addresses, which can be added
		// after the cluster starts. Alternatively we would need to make internal connections to <ip>.ip.local and
		// have coredns convert the IP addresses for us as a DNS query.
		c.ProxyEnv.Upsert(k, strings.Join(append(staticNoProxyRules, v), ","))
		return
	}

	// If we're unable to locate a NO_PROXY config, create default settings
	c.ProxyEnv.Upsert("NO_PROXY", strings.Join(staticNoProxyRules, ","))
}

func moveEntries(dst, src *box.EnvVars, keys ...string) {
	for _, key := range keys {
		deleted := src.Delete(key)
		if deleted != "" {
			dst.Upsert(key, deleted)
		}
	}
}
```

knisbet: I think for the backport it's better to be consistent, and we can refactor the master branch if we want to.

PR #429, opened 2019-06-07
Satellite checks fixes & updates.

r0mant:
This PR:

* Enables time drift check from https://github.com/gravitational/satellite/pull/108.
* Enables ping (latency) check. **Note:** Even though I've enabled it, I discovered a few issues when playing with it. For example, it seems to run but its failure is not visible anywhere except the logs and it does not currently lead to degraded status. I plan to address these issues separately to keep scope of this PR focused.
* Fixes a couple of issues with DNS check by making sure that it uses proper local resolver to resolve internal names.


Gopkg.toml
```diff
@@ -88,7 +88,8 @@ ignored = ["github.com/Sirupsen/logrus", "github.com/gravitational/planet/build/
 
 [[constraint]]
   name = "github.com/gravitational/satellite"
-  version = "=6.0.0"
+  branch = "roman/checks"
+  # version = "=6.0.0"
```
knisbet: Reminder because I would forget: don't forget to revert this.

lib/monitoring/checkers.go
```diff
@@ -178,8 +181,8 @@ func addToMaster(node agent.Agent, config *Config, etcdConfig *monitoring.ETCDCo
 	}
 	node.AddChecker(NewVersionCollector())
 	node.AddChecker(monitoring.NewDNSChecker([]string{
-		"leader.telekube.local",
-	}))
+		"leader.telekube.local.",
+	}, config.LocalNameservers...))
```
knisbet: I wonder if this should detect / skip the check if config is missing? I think the Checker falls back to loading resolv.conf if the nameservers are empty.

tool/planet/start.go
```diff
@@ -197,6 +197,18 @@ func start(config *Config, monitorc chan<- bool) (*runtimeContext, error) {
 			Val:  strings.Join(upstreamNameservers, ","),
 		})
 
+	// Add address of local nameserver to environment.
```
knisbet: Nit: comment should answer why. 
```
// Add address of local nameserver to environment so that processes launched within planet such as planet-agent can locate the local nameserver for testing.
```

a-palchikov:
@r0mant  Regarding the latency check - I thought it was implemented with logging instead of failing in mind (at least the initial version) - so I understood @eldios.

tool/planet/constants.go
```diff
@@ -138,6 +139,9 @@ const (
 	// EnvDNSUpstreamNameservers names the environment variable that specifies
 	// additional nameservers to add to the container's CoreDNS configuration
 	EnvDNSUpstreamNameservers = "PLANET_DNS_UPSTREAM_NAMESERVERS"
+	// EnvDNSLocalNameservers is the container environment variable that
+	// contains address of the node-local DNS servers.
```
a-palchikov: nit: ... that lists node-local DNS nameservers

tool/planet/main.go
```diff
@@ -162,6 +162,7 @@ func run() error {
 		cagentEtcdKeyFile            = cagent.Flag("etcd-keyfile", "TLS key file used to secure etcd communication").String()
 		cagentElectionEnabled        = Bool(cagent.Flag("election-enabled", "Boolean flag to control if the agent initially starts with election participation on").OverrideDefaultFromEnvar(EnvElectionEnabled))
 		cagentDNSUpstreamNameservers = List(cagent.Flag("nameservers", "List of additional upstream nameservers to add to DNS configuration as a comma-separated list of IPs").OverrideDefaultFromEnvar(EnvDNSUpstreamNameservers))
+		cagentDNSLocalNameservers    = List(cagent.Flag("local-nameservers", "List of addresses the local nameserver is listening on").OverrideDefaultFromEnvar(EnvDNSLocalNameservers).Default(DefaultDNSAddress))
```
a-palchikov: `List of local nameserver addresses`. Otherwise it gives an impression there's only a single nameserver (is there?)
It somewhat contradicts the comment on `EnvDNSLocalNameservers` if there's only one.

PR #430, opened 2019-06-12
etcdctl proxy support

knisbet:
Updates https://github.com/gravitational/gravity.e/issues/4115
Requires Backport

In order to workaround issues with etcdctl when run with http_proxy env variables set:
- Rename binary etcdctl -> etcdctl-cmd
- Create wrapper script as etcdctl, that unsets the possible HTTP_PROXY env variables.
- Update etcdctl3 script to also unset http_proxy env variables
- The hardlink that points to the correct etcdctl now is set of etcdctl-cmd.

These changes can be reverted once we go to an etcd version built with golang 1.12 or later. It looks like etcd has already updated master, but it's not in any release yet: https://github.com/etcd-io/etcd/blob/master/Makefile#L54

I did a very basic install and test with these changes, which seem to work.


jeff-1amstudios:
LGTM at least

PR #432, opened 2019-06-17
Make 'gravity shell' error more explicit

webvictim:
Fixes https://github.com/gravitational/gravity.e/issues/4051

### DIFF
--- a/lib/box/client.go
+++ b/lib/box/client.go
@@ -137,7 +137,7 @@ func dial(socketPath string) (net.Conn, error) {
 	conn, err := net.Dial("unix", socketPath)
 	if err != nil {
 		if isPermissionDenied(err) {
-			return nil, trace.Wrap(err, "Permission denied accessing planet socket. Try `sudo gravity shell`")
+			return nil, trace.Wrap(err, "Permission denied accessing planet socket. Try using `sudo`?")
 		}
 		return nil, checkError(trace.Wrap(err, "Failed to connect to planet socket. Check that planet is running."))
 	}
--- a/lib/box/client.go
+++ b/lib/box/client.go
@@ -26,7 +26,6 @@ package box
 import (
 	"encoding/hex"
 	"encoding/json"
-	"fmt"
 	"io"
 	"net"
 	"net/url"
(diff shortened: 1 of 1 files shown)


lib/box/client.go
```diff
@@ -136,7 +136,7 @@ func dial(socketPath string) (net.Conn, error) {
 	conn, err := net.Dial("unix", socketPath)
 	if err != nil {
 		return nil, checkError(
-			trace.Wrap(err, "failed to connect to planet socket"))
+			trace.Wrap(err, "failed to connect to planet. Check that planet is running. Try `sudo gravity shell`?"))
```
a-palchikov: This could be more helpful and detect permission issue by accessing the socket file or interpreting the error from `net.Dial` and only suggesting `sudo` in this case.
Also, there's no reason to make user check whether the container is running if the error is a permission issue. Once that's been sorted out, it could fall back to this suggestion if connection still fails.

lib/box/client.go
```diff
@@ -135,8 +136,19 @@ func (err ExitError) Error() string {
 func dial(socketPath string) (net.Conn, error) {
 	conn, err := net.Dial("unix", socketPath)
 	if err != nil {
```
webvictim: I can't see an easy Go-like way to check whether the error returned by net.Dial contains "permission denied" or not - do you have any advice?

webvictim: @a-palchikov I can't see an easy Go-like way to check whether the error returned by net.Dial contains "permission denied" or not - do you have any advice?

a-palchikov: `net.Dial` wraps the original error into `net.OpError`, so you could use a function similar to this to check whether the error indicates a permission issue:
```go
func dial(socketPath string) {
	conn, err := net.Dial("unix", socketPath)
	if err == nil {
		return conn, nil
	}
	if isPermissionDenied(err) {
		return nil, trace.Wrap(err, "permission denied accessing planet socket. Try with `sudo`?")
	}
	return nil, checkError(trace.Wrap(err, "failed to connect to planet socket"))
}

func isPermissionDenied(err error) bool {
	if opErr, ok := trace.Unwrap(err).(*net.OpError); ok {
		return opErr.Err != nil && os.IsPermission(opErr.Err)
	}
	return os.IsPermission(trace.Unwrap(err))
}
```

webvictim: Thanks!

lib/box/client.go
```diff
@@ -136,19 +136,10 @@ func (err ExitError) Error() string {
...
-		if openErr != nil {
-			if os.IsPermission(openErr) {
-				// socketPath exists but we don't have permission to access it
-				return nil, checkError(
-					trace.Wrap(err, "Permission denied connecting to planet socket. Try `sudo gravity shell`"))
-			}
-			return nil, checkError(
-				trace.Wrap(err, "Unknown error connecting to planet socket"))
+		if isPermissionDenied(err) {
+			return nil, trace.Wrap(err, "Permission denied accessing planet socket. Try `sudo gravity shell`")
```
a-palchikov: nit: I'd avoid referring to a specific command in the error message as it might not be the original command. This method simply does not have the context to deduce this - simply offer to run the command (whatever it was) with `sudo`.

webvictim:
@a-palchikov Thank you for all your reviews and assistance.

PR #441, opened 2019-07-16
Fix planet builds

UnderGreen:


a-palchikov:
I took over these changes in my branch: https://github.com/gravitational/planet/pull/439

PR #442, opened 2019-07-24
Add node problem detector.

r0mant:
This PR adds [Node Problem Detector](https://github.com/kubernetes/node-problem-detector) to Planet. NPD runs as a systemd unit on every node, monitors system logs and sets appropriate conditions on the node it's running node according to its configuration.

Also this PR enables new version of Satellite which monitors node conditions (including those set by NPD) using its `node-status` checker (see https://github.com/gravitational/satellite/pull/112).

Finally, I made a small UX change to pretty-print `planet status` by default.

Refs https://github.com/gravitational/gravity.e/issues/4058.


build.assets/docker/os-rootfs/etc/node-problem-detector/docker-monitor.json
```diff
@@ -0,0 +1,18 @@
+{
+  "plugin": "journald",
+  "pluginConfig": {
+    "source": "dockerd"
+  },
+  "logPath": "/var/log/journal",
+  "lookback": "5m",
+  "bufferSize": 10,
+  "source": "docker-monitor",
+  "conditions": [],
```
a-palchikov: nit: is empty conditions still a requirement for specification?

build.assets/docker/os-rootfs/etc/node-problem-detector/kernel-monitor.json
```diff
@@ -0,0 +1,64 @@
...
+  "rules": [
+    {
+      "type": "temporary",
+      "reason": "OOMKilling",
+      "pattern": "Kill process \\d+ (.+) score \\d+ or sacrifice child\\nKilled process \\d+ (.+) total-vm:\\d+kB, anon-rss:\\d+kB, file-rss:\\d+kB.*"
+    },
+    {
+      "type": "temporary",
+      "reason": "TaskHung",
+      "pattern": "task \\S+:\\w+ blocked for more than \\w+ seconds\\."
```
a-palchikov: nit: the dot at the end is probably not necessary in the pattern

build.assets/docker/os-rootfs/etc/node-problem-detector/systemd-monitor-counter.json
```diff
@@ -0,0 +1,72 @@
...
+      "type": "permanent",
+      "condition": "FrequentContainerdRestart",
+      "reason": "FrequentContainerdRestart",
+      "path": "/usr/bin/log-counter",
+      "args": [
+        "--journald-source=systemd",
+        "--log-path=/var/log/journal",
+        "--lookback=20m",
+        "--count=5",
+        "--pattern=Starting containerd container runtime"
```
a-palchikov: Where does this service come from?

build.assets/docker/os-rootfs/etc/node-problem-detector/docker-monitor.json
r0mant: No idea, this is default configuration. I don't think it hurts anybody.

build.assets/docker/os-rootfs/etc/node-problem-detector/kernel-monitor.json
r0mant: I think it is necessary. IIUC how NPD's matching works, it has to match the entire strings, beginning-to-end. Again, this is the default configuration from their repo.

build.assets/docker/os-rootfs/etc/node-problem-detector/systemd-monitor-counter.json
r0mant: Yeah, it was a part of the default config. I have removed it.

build.assets/docker/os-rootfs/etc/node-problem-detector/docker-monitor.json
r0mant: The checks/conditions are from the "default" configs but I had to tweak some paths for the planet environment and remove some checks not applicable to us. I got a feeling that their "default" [config](https://github.com/kubernetes/node-problem-detector/tree/master/config) is more like a "sample" config which you need to adjust to your environment anyway. I guess if/when we update NPD itself we can also look if they added new stuff to their default config. On the upside, it makes it real easy for us to extend it with our own stuff should we need/want to.

PR #445, opened 2019-08-01
[6.1] Update Kubernetes to 1.15.1.

r0mant:


lib/box/srv.go
```diff
@@ -332,6 +332,12 @@ func getLibcontainerConfig(containerID, rootfs string, cfg Config) (*configs.Con
 				Destination: "/dev/disk",
 				Flags:       syscall.MS_BIND,
 			},
+			{
```
a-palchikov: I understand this is to update to latest satellite that embeds the node problem detector check? But as far as I could tell there was no corresponding backport of node problem detector changes added [here](https://github.com/gravitational/planet/commit/e366d3045b8b4b073f486795408e88e27a073728) or am I missing something?

r0mant: No, node problem detector is not a part of 6.0/6.1. `/dev/kmsg` is now actually required by kubelet as well. I'll add a comment here.

PR #446, opened 2019-08-01
etcd gateway tracking

knisbet:
Fixes https://github.com/gravitational/gravity/issues/535
Requires Backport

This is a quick PR that will stand up a goroutine in planet-agent that will monitor the etcd cluster members list, and keep the etcd gateway upstream list updated with cluster changes and restart the etcd gateway on change.

### DIFF
--- a/tool/planet/etcd.go
+++ b/tool/planet/etcd.go
@@ -287,7 +287,7 @@ func etcdUpgrade(rollback bool) error {
 	return nil
 }
 
-func startWatchingEtcdMasters() error {
+func startWatchingEtcdMasters(ctx context.Context) error {
 	config := etcdconf.Config{
 		Endpoints: []string{DefaultEtcdEndpoints},
 		KeyFile:   DefaultEtcdctlKeyFile,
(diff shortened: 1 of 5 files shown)


tool/planet/etcd.go
```diff
@@ -282,6 +287,91 @@ func etcdUpgrade(rollback bool) error {
...
+				log.WithError(err).Warn("Error resyncing etcd master list.")
+			}
+		case <-ctx.Done():
+			return
+		}
+	}
+}
+
+type etcdGateway struct {
+	clientURLs []string
```
a-palchikov: You don't really need a struct to keep track of client URLs - it could be as simple as a local variable in `watchEtcdMasters`:
```go
func watchEtcdMasters(ctx context.Context, client *etcdv3.Client) {
	ticker := time.NewTicker(15 * time.Second)
	defer ticker.Stop()

	var err error
	var clientURLs []string
	for {
		select {
		case <-ticker.C:
			clientURLs, err = resyncEtcdMasters(ctx, client, clientURLs)
			if err != nil {
				log.WithError(err).Warn("Error resyncing etcd master list.")
			}
		case <-ctx.Done():
			return
		}
	}
}

func resyncEtcdMasters(ctx context.Context, client *etcdv3.Client, clientURLs []string) (newClientURLs []string, err error) {
	memberList, err := client.MemberList(ctx)
	if err != nil {
		return clientURLs, trace.Wrap(err, "error retrieving member list")
	}

	newClientURLs, err := collectClientURLs(memberList)
	if err != nil {
		return clientURLs, trace.Wrap(err)
	}

	// Only rewrite the configuration if there are changes
	sort.Strings(newClientURLs)
	if reflect.DeepEqual(newClientURLs, clientURLs) {
		return clientURLs, nil
	}

	err = updateEtcdClientURLs(newClientURLs)
	if err != nil {
		return clientURLs, trace.Wrap(err)
	}

	return newClientURLs, nil
}
```

tool/planet/etcd.go
```diff
@@ -282,6 +287,91 @@ func etcdUpgrade(rollback bool) error {
...
+	sort.Strings(newClientURLs)
+	if reflect.DeepEqual(newClientURLs, e.clientURLs) {
+		return nil
+	}
+
+	env := fmt.Sprintf("%v=%q", EnvEtcdGatewayEndpoints, strings.Join(newClientURLs, ","))
+	log.WithField("file", DefaultEtcdSyncedEnvFile).Info("Updating etcd gateway environment: ", env)
+	err = utils.SafeWriteFile(DefaultEtcdSyncedEnvFile, []byte(env), constants.SharedReadMask)
+	if err != nil {
+		return trace.Wrap(err, "failed to update etcd synced environment").AddField("file", DefaultEtcdSyncedEnvFile)
```
a-palchikov: `failed to sync etcd environment`?

tool/planet/etcd.go
```diff
@@ -282,6 +287,91 @@ func etcdUpgrade(rollback bool) error {
...
+	}
+
+	err = systemctl(ctx, "restart", ETCDServiceName)
```
a-palchikov: There's also a concept of a [path](https://www.freedesktop.org/software/systemd/man/systemd.path.html#Unit=) systemd unit which could be used to track changes to etcd-relevant files - i.e. it could watch other environment files and restart etcd on changes - it could be more consistent to do that way.

tool/planet/etcd.go
knisbet: I'm not sure I see how this makes it any simpler?

tool/planet/etcd.go
```diff
@@ -282,6 +287,91 @@ func etcdUpgrade(rollback bool) error {
...
+
+	go watchEtcdMasters(ctx, cli)
+	return nil
+}
+
+func watchEtcdMasters(ctx context.Context, client *etcdv3.Client) {
+	ticker := time.NewTicker(15 * time.Second)
+	defer ticker.Stop()
+
+	gateway := etcdGateway{}
```
r0mant: IIUC because this is initialized with an empty `clientURLs`, it will lead to the etcd-gateway restart right away because it will detect the members below and compare them against an empty list.

I would initialize this with current URLs instead. Otherwise, let's say you restart planet-agent and it will in turn trigger etcd-gateway restart.

tool/planet/etcd.go
r0mant: I don't have a strong opinion on using a struct vs a local var here but I do like Dima's suggestion on encapsulating some logic into `collectClientURLs(memberList)` and `updateEtcdClientURLs(newClientURLs)` functions.

tool/planet/etcd.go
```diff
@@ -282,6 +288,97 @@ func etcdUpgrade(rollback bool) error {
...
+
+	e.clientURLs = newClientURLs
+	return nil
+}
+
+func collectClientURLs(memberList *etcdv3.MemberListResponse) ([]string, error) {
+	newClientURLs := []string{}
+	for _, member := range memberList.Members {
+		memberURLs := member.GetClientURLs()
+		if len(memberURLs) == 0 {
```
anandkumarpatel: @knisbet any chance we can actually make this error if `< 3` ?   not sure if less than 2 etcd servers is valid. (1 is the address of this node)

knisbet: This doesn't protect against there being less than 2 etcd servers. This protects against the unexpected situation where we encounter a cluster member that doesn't have any client urls defined on the member. 

On a different note, it's quite common to run a single node cluster among other deployments, that don't have redundancy.

PR #469, opened 2019-08-21
dynamic detection of the entire set of linux capabilities

knisbet:
CAP_AUDIT_READ isn't available on all supported kernels, so this copies the libcontainer behaviour to detect the capabilities offered by the kernel and add support. 

Note: Manual tests still pending.

### DIFF
--- a/tool/planet/constants.go
+++ b/tool/planet/constants.go
@@ -18,10 +18,12 @@ package main
 
 import (
 	"fmt"
+	"strings"
 	"time"
 
 	"github.com/gravitational/planet/lib/utils"
 
+	"github.com/syndtr/gocapability/capability"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	kubeletconfig "k8s.io/kubelet/config/v1beta1"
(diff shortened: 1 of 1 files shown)


tool/planet/constants.go
```diff
@@ -542,43 +544,21 @@ var KubeletVersion = schema.GroupVersionKind{
...
-	"CAP_SYSLOG",
-	"CAP_WAKE_ALARM",
+var allCaps = []string{}
+
+// Based on runc capabilities detection
+// https://github.com/opencontainers/runc/blob/2e94378464ae22b92e1335c200edb37ebc94a1b7/libcontainer/capabilities_linux.go#L17-L31
+func init() {
+	last := capability.CAP_LAST_CAP
+	// workaround for RHEL6 which has no /proc/sys/kernel/cap_last_cap
+	if last == capability.Cap(63) {
```
a-palchikov: This feels like a sub-optimal way of detecting RHEL6. What if there are 64 capabilities supported? I think it would be safer to probe down from `capabilities.CAP_LAST_CAP` until one is found that works and set that as the last for the case it equals `63` initially - i.e. either it has not been read from cap_last_cap file (as the case is with some RHEL6 distributions) or it has read `63` (in which case it might not be RHEL6) - what do you think?

knisbet: I think we don't support RHEL6 so it's not really worth worrying about, unless there is some evidence cap_last_cap is also missing in RHEL 7.4+. This was literally just a direct copy of the logic from libcontainer but ported to be a list instead of a map which planet is already using.

PR #479, opened 2019-09-09
remove setting a default upstream resolver of 127.0.0.1 when reading …

knisbet:
…resolv.conf settings

Updates https://github.com/gravitational/gravity/issues/613
Requires backport

When reading DNS configuration for planet, if no upstream nameservers are configured, a default of 127.0.0.1 is added as an upstream resolver. It's not clear to me why this was done, but I don't believe it's currently necessary or beneficial to do so.

### DIFF
--- a/lib/utils/dns.go
+++ b/lib/utils/dns.go
@@ -33,8 +33,6 @@ import (
 	"strings"
 )
 
-var defaultNS = []string{"127.0.0.1", "::1"}
-
 type DNSConfig struct {
 	Servers    []string // servers to use
 	Domain     string   // Domain parameter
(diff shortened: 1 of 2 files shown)


a-palchikov:
I believe this was a direct port of Go's DNS configuration reader and the relevant changes that introduced it are [here](https://github.com/golang/go/commit/f390135733ae20a3c8b6700426cc379f6e145319#diff-c73f1d143b49688cf697ab657c6f2988R11) and here's the corresponding [issue](https://github.com/golang/go/issues/10566).

knisbet:
Hmmm ok, that makes it tough. In gravity this change stems from, we don't want to configure a non-existant forwarder in coredns, and presumably anywhere else we use the resolv.conf to a potentially non-existant DNS server. The difficulty is, knowing this, it seems a blank resolv.conf can mean I don't have any resolvers, or it can mean I have a localhost resolver. 

My knee jerk reaction, is to treat an empty resolv.conf as a I have no upstream resolvers, but there is the potential that this view bites us.

a-palchikov:
Here's the [implementation](https://github.com/containers/libpod/blob/v1.6.3/pkg/resolvconf/resolvconf.go#L105) of resolv.conf set up for containers performed by libpod - might be worth a look.

knisbet:
Closed in favor of #731

PR #482, opened 2019-09-16
restart etcd clients within planet during etcd upgrade

knisbet:
Still testing

Requires backports up to 5.2.x.

Resolves an issue where the etcd clients get confused after the upgrade, due to the cluster being re-created, but active watches can point to a revision that is in the future, causing watches not to fire. All clients of etcd need to be restarted to avoid this issue when the etcd cluster is re-created.

### DIFF
--- a/tool/planet/etcd.go
+++ b/tool/planet/etcd.go
@@ -170,6 +170,8 @@ func etcdEnable(upgradeService bool) error {
 	defer cancel()
 
 	if !upgradeService {
+		// restart the clients of the etcd service when the etcd service is brought online, which usually will be post
+		// upgrade. This will ensure clients running inside planet are restarted, which will refresh any local state
 		restartEtcdClients(ctx)
 		return trace.Wrap(enableService(ctx, ETCDServiceName))
 	}
@@ -197,7 +199,7 @@ func etcdUpgrade(rollback bool) error {
 	}
 
 	if env.Get(EnvEtcdProxy) == EtcdProxyOn {
-		log.Info("etcd is in proxy mode, restarting etcd clients")
+		log.Info("etcd is in proxy mode, nothing to do")
 		return nil
 	}
 
@@ -284,9 +286,12 @@ func restartEtcdClients(ctx context.Context) {
 	for _, service := range services {
 		// reset the kubernetes api server to take advantage of any new etcd settings that may have changed
 		// this only happens if the service is already running
-		status, err := getServiceStatus(APIServerServiceName)
+		status, err := getServiceStatus(service)
 		if err != nil {
-			log.Warnf("Failed to query status of service %v. Continuing upgrade. Error: %v", APIServerServiceName, err)
+			log.WithFields(log.Fields{
+				log.ErrorKey: err,
+				"service":    service,
+			}).Warn("Failed to query service status.")
 			return
 		}
 		if status != "inactive" {
--- a/tool/planet/etcd.go
+++ b/tool/planet/etcd.go
@@ -170,6 +170,7 @@ func etcdEnable(upgradeService bool) error {
 	defer cancel()
 
 	if !upgradeService {
+		restartEtcdClients(ctx)
 		return trace.Wrap(enableService(ctx, ETCDServiceName))
 	}
 	// don't actually enable the service if this is a proxy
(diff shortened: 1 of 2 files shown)


tool/planet/etcd.go
```diff
@@ -272,22 +271,30 @@ func etcdUpgrade(rollback bool) error {
...
 
+// restartEtcdClients - because the etcd cluster has been recreated, all clients need to be refreshed so their
+// watches are not pointing at incorrect revisions.
+func restartEtcdClients(ctx context.Context) {
+	services := []string{APIServerServiceName, PlanetAgentServiceName, FlannelServiceName}
+
+	for _, service := range services {
+		// reset the kubernetes api server to take advantage of any new etcd settings that may have changed
+		// this only happens if the service is already running
+		status, err := getServiceStatus(APIServerServiceName)
```
r0mant: Should not this be `getServiceStatus(service)`? Ditto for the warning message below.

tool/planet/etcd.go
```diff
@@ -170,6 +170,7 @@ func etcdEnable(upgradeService bool) error {
 	defer cancel()
 
 	if !upgradeService {
+		restartEtcdClients(ctx)
```
a-palchikov: To clarify this change - when the etcd service (not upgrade service!) is re-enabled (being the last step), the clients are restarted?

tool/planet/etcd.go
```diff
@@ -198,7 +197,7 @@ func etcdUpgrade(rollback bool) error {
 	}
 
 	if env.Get(EnvEtcdProxy) == EtcdProxyOn {
-		log.Info("etcd is in proxy mode, nothing to do")
+		log.Info("etcd is in proxy mode, restarting etcd clients")
```
a-palchikov: Why is the message about restarting clients here and not in the method that does the restart?

tool/planet/etcd.go
```diff
@@ -272,22 +271,30 @@ func etcdUpgrade(rollback bool) error {
...
+// watches are not pointing at incorrect revisions.
+func restartEtcdClients(ctx context.Context) {
+	services := []string{APIServerServiceName, PlanetAgentServiceName, FlannelServiceName}
+
+	for _, service := range services {
+		// reset the kubernetes api server to take advantage of any new etcd settings that may have changed
+		// this only happens if the service is already running
+		status, err := getServiceStatus(APIServerServiceName)
+		if err != nil {
+			log.Warnf("Failed to query status of service %v. Continuing upgrade. Error: %v", APIServerServiceName, err)
```
a-palchikov: nit: put the service name into separate field:
```go
log.WithFields(log.Fields{
	log.ErrorKey: err,
	"service":    service,
}).Warn("Failed to query service status.")
```

tool/planet/etcd.go
knisbet: correct. The upgrade service is a temp cluster that external clients can't reach, that is used to restore the backups. Then it get's restarted back on the regular ports, and this is when we trigger the restart.

tool/planet/etcd.go
knisbet: my bad, I originally added it the restart here, but then changed my mind and linked it into the service start code.

tool/planet/etcd.go
```diff
@@ -272,22 +273,33 @@ func etcdUpgrade(rollback bool) error {
...
+		// this only happens if the service is already running
+		status, err := getServiceStatus(service)
+		if err != nil {
+			log.WithFields(log.Fields{
+				log.ErrorKey: err,
+				"service":    service,
+			}).Warn("Failed to query service status.")
+			return
+		}
+		if status != "inactive" {
```
r0mant: @knisbet @a-palchikov Btw, I've just realized that this `!= "inactive"` logic probably applies to kube-apiserver only, in order to filter out inactive apiservers running on standby masters? It probably should not kick in for other services.

knisbet: it also applies to flannel, since flannel isn't necessarily enabled / running due to support for third party plugins and wormhole. And I don't see it doing harm to do the same thing for gravity-agent. If the process is inactive, why should the etcd upgrade touch it.

PR #489, opened 2019-09-18
5.5.x etcd additional restarts

knisbet:
This is a workaround to issues we've been seeing with upgrades that invoke the etcd upgrade behaviour. It relates to when originally designing the etcd upgrade, the kubernetes docs at the time exposed the underlying etcd index as a resourceVersion, but at the time I believe they indicated that clients should not rely on this behaviour. It looks like since then, kubernetes now does rely on the underlying etcd watch behaviour, which means kubernetes api consumers are now exposed to the same issues with the index resetting during the etcd upgrade.

This is a temporary workaround, to ensure the kubernetes apiserver is shutdown during the etcd upgrade, so that it is not able to serve any stale data to running clients. Each kubernetes client within planet is also restarted at this time. When planet-agent is restarted, planet-agent will be responsible for restarting the kubernetes api based on election results.

After discussing with @a-palchikov we'll have to take a longer term look at this as software running within the cluster using the kubernetes API will be exposed to the same problems.

Symptoms:
- kube-proxy / kubelet are running, but don't appear to be processing updates to objects from the API.

Note: I added stopping the api server as a flag, because I have used the planet command before during troubleshooting to stop etcd, and I believe it's unintuitive that this also stops the apiserver. 

Note: manual testing on a cluster that very frequently shows the kubernetes communications issues did not show symptoms after this workaround with 2 attempts.

### DIFF
--- a/tool/planet/constants.go
+++ b/tool/planet/constants.go
@@ -317,10 +317,16 @@ const (
 	ETCDUpgradeServiceName = "etcd-upgrade.service"
 	// APIServerServiceName names the service unit for k8s apiserver
 	APIServerServiceName = "kube-apiserver.service"
+	// ProxyServiceName is the name of the k8s proxy systemd service
+	ProxyServiceName = "kube-proxy.service"
+	// KubeletServiceName is the name of the k8s kubelet systemd service
+	KubeletServiceName = "kube-kubelet.service"
 	// PlanetAgentServiceName is the name of the planet agent
 	PlanetAgentServiceName = "planet-agent.service"
 	// FlannelServiceName is the name of the flannel service
 	FlannelServiceName = "flanneld.service"
+	// CorednsServiceName is the name of the flannel service
+	CorednsServiceName = "coredns.service"
 
 	// ETCDDropinPath is the location of the systemd dropin when etcd is in gateway mode
 	ETCDGatewayDropinPath = "/etc/systemd/system/etcd.service.d/10-gateway.conf"
--- a/tool/planet/etcd.go
+++ b/tool/planet/etcd.go
@@ -153,13 +153,24 @@ func etcdBackup(backupFile string) error {
 
 // etcdDisable disables etcd on this machine
 // Used during upgrades
-func etcdDisable(upgradeService bool) error {
+func etcdDisable(upgradeService, stopAPIServer bool) error {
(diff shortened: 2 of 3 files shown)


tool/planet/constants.go
```diff
@@ -317,10 +317,16 @@ const (
...
 	APIServerServiceName = "kube-apiserver.service"
+	// ProxyServiceName is the name of the k8s proxy systemd service
+	ProxyServiceName = "kube-proxy.service"
+	// KubeletServiceName is the name of the k8s kubelet systemd service
+	KubeletServiceName = "kube-kubelet.service"
 	// PlanetAgentServiceName is the name of the planet agent
 	PlanetAgentServiceName = "planet-agent.service"
 	// FlannelServiceName is the name of the flannel service
 	FlannelServiceName = "flanneld.service"
+	// CorednsServiceName is the name of the flannel service
```
r0mant: flannel -> coredns

tool/planet/etcd.go
```diff
@@ -153,13 +153,24 @@ func etcdBackup(backupFile string) error {
...
-func etcdDisable(upgradeService bool) error {
+func etcdDisable(upgradeService, stopAPIServer bool) error {
 	ctx, cancel := context.WithTimeout(context.Background(), EtcdUpgradeTimeout)
 	defer cancel()
 
+	// Kevin: Workaround, for the API server presenting stale data to clients while etcd is down. Make sure we shut down
+	// the API server as well (passed as flag from gravity to prevent accidental usage).
+	// TODO: This fix needs to be revisited to include a permanent solution.
+	if stopAPIServer {
+		err := systemctl(ctx, "stop", APIServerServiceName)
```
r0mant: Is this going to return an error if the apiserver is already stopped (e.g. on non-active leader) node, or it won't?

knisbet: Doesn't seem to:
```
kevin-test4:/$ systemctl stop kube-apiserver
kevin-test4:/$ echo $?
0
kevin-test4:/$ systemctl stop kube-apiserver
kevin-test4:/$ echo $?
0
```

knisbet: I don't think so:
```
kevin-test4:/$ systemctl stop kube-apiserver
kevin-test4:/$ echo $?
0
kevin-test4:/$ systemctl stop kube-apiserver
kevin-test4:/$ echo $?
0
```

PR #526, opened 2019-11-04
disable automatic node registration

knisbet:
Remove automatic node rotation from kubelet.

Updates https://github.com/gravitational/gravity/issues/863


build.assets/makefiles/master/k8s-master/kube-kubelet.service
```diff
@@ -14,8 +14,7 @@ ExecStart=/usr/bin/kubelet \
         --cluster-dns=${DNS_ADDRESSES} \
         --pod-infra-container-image=${KUBE_APISERVER}:5000/gcr.io/google_containers/pause:3.0 \
         --kubeconfig=/etc/kubernetes/kubelet.kubeconfig \
-        --register-with-taints=${PLANET_NODE_TAINTS} \
```
a-palchikov: Do we still need these environment variables? If not, maybe remove them as well.

PR #555, opened 2020-02-07
[6.1.x] Integrate Satellite Timeline

bernardjkim:
- Revendor satellite v6.1.3
- Update agent config


a-palchikov:
Without changes to the build environment the planet won't start on redhat 7.7:
```
rootfs/usr/bin/planet: /lib64/libc.so.6: version `GLIBC_2.28' not found (required by rootfs/usr/bin/planet)
```
I remember you mentioned encountering this in your tests (or during development) - were you able to work around it? If not, we need to look into this first.

bernardjkim:
Ah, I only made the changes to the build environment when building gravity. We decided to use `golang1.12.9-stretch` for the build env instead of static linking musl.
I think changing the BUILDBOX_GO_VER from `1.12.9` -> `1.12.9-stretch` in planet makefile should fix this.

a-palchikov:
No, `stretch` should be fine since there's just `fcntl` API that's a breaking change in `2.28` which is used by the runc and it'd fallback to the version from `2.2.5` on `stretch`. Since planet (incl. indirectly) is not using `fcntl64` with support for large files (at least not that I know of), it should be fine.

Cannot think of any downsides at the moment and also see no easy alternative as we need to support systems with `GLIBC` (much) lower than `2.28`.

PR #557, opened 2020-02-11
Updated Kubernetes to 1.17.3 and revendored deps

eldios:
Updated Kubernetes to 1.17.2 and revendored deps


eldios:
1.17.3 came out, so switching to that instead

PR #560, opened 2020-02-24
SELinux support

a-palchikov:


build.assets/makefiles/master/k8s-master/kube-kubelet.service
```diff
@@ -9,7 +9,7 @@ EnvironmentFile=/run/dns.env
 EnvironmentFile=-/run/kubernetes-reserved.env
 ExecStartPre=/bin/systemctl is-active docker.service
 ExecStart=/usr/bin/kubelet \
-        --root-dir=/var/lib/gravity/planet/kubelet \
+        --root-dir=/var/lib/kubelet \
```
r0mant: Didn't we move this directory under /var/lib/gravity?

a-palchikov: Yes, we did but I wanted to keep this explicitly under `/var/lib/kubelet` (as this is the default location in kubelet source code) due to its unfortunate hard-coding for some of the paths which then interfere when SELinux is enabled - hence `/var/lib/gravity/planet/kubelet` won't work.

r0mant (approved):

lib/box/enter.go
```diff
@@ -48,14 +50,25 @@ func (err ExitError) Error() string {
...
-func CombinedOutput(dataDir string, cfg *ProcessConfig) ([]byte, error) {
-	var b bytes.Buffer
-	cfg.Out = &b
-	err := Enter(dataDir, cfg)
+func CombinedOutput(config EnterConfig) ([]byte, error) {
+	var buf bytes.Buffer
+	config.Process.Out = &buf
+	err := Enter(config)
+	if err != nil {
+		return buf.Bytes(), err
```
r0mant: trace.Wrap?

lib/box/enter.go
```diff
@@ -48,14 +50,25 @@ func (err ExitError) Error() string {
...
+}
+
+// CombinedOutput runs a process within planet, returning the output as a byte buffer
+func (b *Box) CombinedOutput(config ProcessConfig) ([]byte, error) {
+	var buf bytes.Buffer
+	config.Out = &buf
+	err := enter(b.dataDir, b.Container, config)
 	if err != nil {
-		return b.Bytes(), err
+		return buf.Bytes(), err
```
r0mant: trace.Wrap?

lib/box/enter.go
```diff
@@ -229,38 +207,107 @@ func Enter(dataDir string, cfg *ProcessConfig) error {
...
+	absRoot, err := filepath.Abs(dataDir)
+	if err != nil {
+		return nil, trace.Wrap(err)
+	}
+	files, err := ioutil.ReadDir(absRoot)
+	if err != nil {
+		return nil, trace.Wrap(err)
+	}
+	if len(files) == 0 {
+		return nil, trace.NotFound("planet container not found").AddField("data_dir", dataDir)
```
r0mant: Nit: I think we usually use dashes for multi-word fields, so I'd call this one "data-dir".

lib/box/enter.go
```diff
@@ -229,38 +207,107 @@ func Enter(dataDir string, cfg *ProcessConfig) error {
...
-			return nil, err
+			return nil, trace.Wrap(err)
 		}
 
-		process.ConsoleSocket = child
-		t.postStart = append(t.postStart, parent, child)
-		t.consoleC = make(chan error, 1)
+		// There should only be a single planet container that's running, so exec within the first
+		// running container found
+		if status == libcontainer.Running {
```
r0mant: Should we log something here, i.e. to indicate which container we're entering, in case there're multiple running (which shouldn't happen under normal circumstances but we've seen this happen)?

lib/box/enter.go
```diff
@@ -229,38 +207,107 @@ func Enter(dataDir string, cfg *ProcessConfig) error {
...
+			break
+		}
+	}
+	if status == libcontainer.Stopped {
+		return nil, trace.BadParameter("cannot exec into stopped container").AddField("container", container.ID())
```
r0mant: Is user going to potentially see this error? If so, we should make this message more user-friendly, something like "planet container is stopped, please start it using ...".

lib/box/selinux.go
```diff
@@ -28,24 +28,20 @@ import (
 	log "github.com/sirupsen/logrus"
 )
 
-// getSELinuxProcLabel returns a new function to compute SELinux labels.
-//
-// The returned function will compute the appropriate SELinux domain for the
+// getSELinuxProcLabel computes the appropriate SELinux domain for the
 // command specified with cmd and fall back to defaults.ContainerProcessLabel
```
r0mant: Nit: fall -> falls.

tool/planet/udev.go
```diff
@@ -112,30 +114,30 @@ func (r *udevListener) createDevice(device *configs.Device) error {
 		return trace.Wrap(err)
 	}
 
-	err = enter(deviceCmd("add", "--data", string(deviceJson)))
+	err = enter(deviceCmd(r.seLinux, "add", "--data", string(deviceJson)))
```
r0mant: Nit: I'd just make `deviceCmd` a udevListener method so you don't have to pass this new parameter.

tool/planet/udev.go
```diff
@@ -32,7 +32,7 @@ import (
 
 // newUdevListener creates a new udev event listener listening
 // for events on block devices of type `disk`
-func newUdevListener() (*udevListener, error) {
+func newUdevListener(seLinux bool) (*udevListener, error) {
```
r0mant: Random thought: I wonder if you considered making this a global parameter somewhere that would be initialized e.g. in an init() method, rather than having to pass it everywhere.

a-palchikov: Not sure I'd fancy anything like that - I think it's better when it's explicit even if a bit repetitive.

PR #563, opened 2020-02-26
Enable more planet agent logging.

r0mant:
Enable verbose logging for planet agent. We've recently had cases where planet-agent would be down or not reacting to leader change events and we had literally zero info to tell why.

Right now it is pretty much impossible to tell what's happening with the planet agent b/c in the normal mode of operations it doesn't write any logs at all. Occasionally it logs warnings but it's impossible to, for example, see what's happening with its leader election and the agent being silent gives the impression that it's stuck. I have a suspicion that the watch in coordinate library gets broken somehow but can't confirm or deny this theory without having any logs (coordinate library logs lots of stuff at debug level).

Let me know what you think. I'm for just having it enabled at all times as it doesn't really produce a lot of logs (compared to other services at least) and will at least give us some insight into troubleshooting issues.

Needs ports to 6.1, 6.3, master.


build.assets/makefiles/base/agent/planet-agent.service
```diff
@@ -14,7 +14,7 @@ LimitNOFILE=40000
 LimitNPROC=1048576
 ExecStartPre=/bin/systemctl is-active etcd.service
 ExecStartPre=/bin/systemctl is-active serf.service
-ExecStart=/usr/bin/planet agent \
+ExecStart=/usr/bin/planet agent --debug \
```
a-palchikov: this will just spam the logs won't it? Shouldn't this instead set the level to info and focus on more informative logs at info?

knisbet:
I'm a bit on the fence. Checking on my test cluster, the volume of logs doesn't seem that bad, but definitely has some noise to them as well. In general I think info logs would be more valuable, but I don't know that the info logs would really contain any insights, although if we're not generating errors, the debug logs might not provide much insight either.

r0mant:
@a-palchikov @knisbet That's fair. It has info level by default but the coordinate library logs everything at debug for some reason, even errors. I agree a better approach is to fix inadequate logging in coordinate, I'll do that instead.

PR #564, opened 2020-02-26
preserve etcd revision during ugrades

knisbet:
This PR definitely scope creeped as I went through it:
- Updates etcd to 3.4.x
- I rewrote the container enter code, so it's no longer based on the websocket/planet.sock connectivity through a webserver. It's largely a port of the runc exec code, although I've tweaked it a bit to be planet specific, and skipped all the logic for non-root containers, etc. Unfortunately it's not exported code, so it ended up be largely a port. This was due to our pty handling having some problems dealing with terminal widths after some dependency updates.
- Etcd 3.4 seems to be more strict when reconnecting the cluster, so I added some join logic so that node can add itself to the cluster then join immediately as a single step. It's also alot more strict with superflous flags, so I had to do some cleanup on the flags we're passing.
- I've introduced mage here as well with a single target, to make it easier to write an integration test for the etcd upgrade. This runs a single node test, that sets up watches, and verifies the behaviour when deleting the database and restoring with a new database.
- Some additional tweaks to the etcd DB restoration, where the planet code is responsible for stopping/starting the temporary etcd service, instead of being controlled by gravity.

Requires: https://github.com/gravitational/etcd-backup/pull/4
Requires: https://github.com/gravitational/coordinate/pull/11


a-palchikov (commented):
I was not able to look through all of the changes yet - will resume tomorrow.

lib/box/signal.go
```diff
@@ -0,0 +1,125 @@
...
+
+	for signal := range s.signalC {
+		switch signal {
+		case unix.SIGWINCH:
+			// Ignore errors resizing, as above.
+			_ = tty.resize()
+		case unix.SIGCHLD:
+			exits, err := s.reap()
+			if err != nil {
+				logrus.WithError(err).Error("error repeating child")
```
r0mant: reaping

lib/box/tty.go
```diff
@@ -0,0 +1,164 @@
+/*
+Copyright 2020 Gravitational, Inc.
```
r0mant: Is this file a complete copy-paste from runc source? If so, then we should probably use their license header.

tool/planet/etcd.go
```diff
@@ -196,9 +200,97 @@ func etcdEnable(upgradeService bool) error {
 		log.Info("etcd is in proxy mode, nothing to do")
 		return nil
 	}
+
 	return trace.Wrap(enableService(ctx, ETCDUpgradeServiceName))
 }
 
+// etcdInitJoin ensures this particular node is part of an etcd cluster.
+// Because the etcd cluster is re-created during an upgrade, this particular node may not be part of the new cluster.
```
r0mant: Could you elaborate on why this may be the case and why we didn't need to do this before?

tool/planet/etcd.go
```diff
@@ -412,23 +504,100 @@ func getBaseEtcdDir(version string) string {
...
+	}
+
+	// To be re-entrant, shutdown the etcd server if it's already running
+	err = etcdDisable(true, false)
+	if err != nil {
+		return trace.Wrap(err)
+	}
+
+	// For the restore operation to be re-entrant, delete the data directory in-case the step is being re-run
+	err = os.RemoveAll(datadir)
```
r0mant: Should it be backed up first?

tool/planet/etcd.go
```diff
@@ -412,23 +504,100 @@ func getBaseEtcdDir(version string) string {
...
+	}
+
+	// wait for the temporary etcd instance to complete startup
+	log.Info("Waiting for etcd initialization to complete")
+	err = waitEtcdHealthyTimeout(context.TODO(), 2*time.Minute, client)
+	if err != nil {
+		return trace.Wrap(err)
+	}
+
+	// stop etcd now that it's DB is initialized but empty, to run offline restore to the empty database
```
r0mant: Nit: it's -> its

tool/planet/etcd.go
```diff
@@ -412,23 +504,100 @@ func getBaseEtcdDir(version string) string {
...
+	log.Info("Etcd initialization complete, stopping")
+	err = etcdDisable(true, false)
```
r0mant: Nit: To help the reader figure out what "true" and "false" are I would either create variables and user them here (i.e. `var confirmed = true, etcdDisable(confirmed, ...)` or whatever they are), or convert them into a struct. Same for `etcdEnable`.

tool/planet/etcd.go
```diff
@@ -196,9 +200,97 @@ func etcdEnable(upgradeService bool) error {
...
+	}
+
+	// Gravity will give us a hint about which master to join if we need to run init
+	if initMaster == "" {
+		return nil
+	}
+
+	log.WithField("master", initMaster).Info("Joining etcd to existing cluster")
+
+	return trace.Wrap(utils.Retry(ctx, math.MaxInt64, 1*time.Second, func() error {
```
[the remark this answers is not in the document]
r0mant: Agree. Not the loop itself though but the function being retried.

tool/planet/etcd.go
```diff
@@ -196,9 +200,97 @@ func etcdEnable(upgradeService bool) error {
...
+			return trace.BadParameter("public ip env variable is empty").AddField(EnvPublicIP, publicIP)
+		}
+
+		advertisePeerUrl := fmt.Sprintf("https://%v:2380", publicIP)
+
+		resp, err := client.MemberList(ctx)
+		if err != nil {
+			return trace.Wrap(err)
+		}
+		peerUrls := []string{}
```
r0mant: Nit: peerURLs

tool/planet/etcd.go
```diff
@@ -196,9 +200,97 @@ func etcdEnable(upgradeService bool) error {
...
+		if err != nil {
+			return trace.Wrap(err)
+		}
+
+		publicIP := env.Get(EnvPublicIP)
+		if len(publicIP) == 0 {
+			return trace.BadParameter("public ip env variable is empty").AddField(EnvPublicIP, publicIP)
+		}
+
+		advertisePeerUrl := fmt.Sprintf("https://%v:2380", publicIP)
```
r0mant: Nit: advertisePeerURL

tool/planet/etcd.go
```diff
@@ -196,9 +200,97 @@ func etcdEnable(upgradeService bool) error {
...
+
+		resp, err := client.MemberList(ctx)
+		if err != nil {
+			return trace.Wrap(err)
+		}
+		peerUrls := []string{}
+
+		isMember := false
+		for _, member := range resp.Members {
```
r0mant: Any way to factor out this logic into a method like `func isMember(client etcd.Client, advertiseIP string) bool`?

lib/box/signal.go
```diff
@@ -0,0 +1,125 @@
...
+		ws  unix.WaitStatus
+		rus unix.Rusage
+	)
+	for {
+		pid, err := unix.Wait4(-1, &ws, unix.WNOHANG, &rus)
+		if err != nil {
+			if err == unix.ECHILD {
+				return exits, nil
+			}
+			return nil, err
```
r0mant: trace.Wrap(err)

build.assets/makefiles/etcd/etcd-upgrade.service
```diff
@@ -18,14 +18,15 @@ EnvironmentFile=-/ext/etcd/etcd-version.txt
...
+UnsetEnvironment=ETCD_INITIAL_CLUSTER_STATE ETCD_INITIAL_CLUSTER
 ExecStartPre=/usr/bin/planet etcd init
 ExecStart=/usr/bin/etcd \
         --name=${PLANET_ETCD_MEMBER_NAME} \
         --data-dir=/ext/etcd/${PLANET_ETCD_VERSION} \
         --initial-advertise-peer-urls=https://${PLANET_PUBLIC_IP}:2380 \
         --advertise-client-urls=https://127.0.0.2:2379,https://127.0.0.2:4001 \
         --listen-client-urls=https://127.0.0.2:2379,https://127.0.0.2:4001 \
-        --listen-peer-urls=https://${PLANET_PUBLIC_IP}:2380,https://${PLANET_PUBLIC_IP}:7001 \
+        --listen-peer-urls=https://127.0.0.2:2380,https://127.0.0.2:7001 \
```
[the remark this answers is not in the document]
knisbet: Yea, to my knowledge we don't need the additional ports, I just sort of left them because I didn't know why it was configured this way in the first place. Last time I went through, I think all client comms are correctly using 2379 only, and server to server comms only need the one port, which 2380 should be fine for.

Would also allow us to simplify the requirements in our docs ever so slightly.

lib/box/enter.go
```diff
@@ -18,195 +18,272 @@ package box
...
-	// with "init" command line argument.  (this is the default setting)
-	// then our init() function comes into play
-	if err := c.Run(p); err != nil {
-		return trace.Wrap(err)
+	// This is a bit of a risk, try and use the cpu controller to identify the cgroup path. CgroupsV1 doesn't use a
+	// unified hierarchy, so different controllers can have different cgroup paths. For us, cpu is the most important
+	// controller, so we'll use it as the reference
+	cgroupPath, ok := state.CgroupPaths["cpu"]
+	if !ok {
+		return trace.NotFound("cpu cgroup controller not found: %v", state.CgroupPaths)
```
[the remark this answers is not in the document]
knisbet: I'm pretty sure this code predates the AddField method. Just in rewriting other sections of code seems to have gotten the diff algorithm confused.

lib/box/enter.go
```diff
@@ -18,195 +18,272 @@ package box
...
+	if !strings.HasPrefix(cgroupPath, "/sys/fs/cgroup/") {
+		return trace.BadParameter("Cgroup path not mounted to /sys/fs/cgroup: %v", cgroupPath)
 	}
-	log.Debugf("Process %#v started.", p)
 
-	setProcessUserCgroup(c, p)
+	// Example cgroup path: /sys/fs/cgroup/cpu,cpuacct/system.slice/-planet-cee2b8a0-c470-44a6-b7cc-1eefbc1cc88c.scope
+	// we want to split off the /sys/fs/cgroup/cpu,cpuacct/ part, so we have just the cgroup structure
+	// (system.slice/-planet-cee2b8a0-c470-44a6-b7cc-1eefbc1cc88c.scope)
+	dirs := strings.Split(cgroupPath, "/")
```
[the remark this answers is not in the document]
knisbet: Again, this is existing code caught in the diff algorithm.

lib/box/enter.go
```diff
@@ -18,195 +18,272 @@ package box
...
 
-	control, err := cgroups.Load(cgroups.V1, cgroups.StaticPath(userPath))
+	if s != 0 {
+		return trace.Wrap(&ExitError{Code: s})
+	}
+	return nil
+}
+
+func setupIO(process *libcontainer.Process, rootuid, rootgid int, createtty bool) (*tty, error) {
+	if createtty {
```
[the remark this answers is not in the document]
knisbet: This is consistent naming with the runc code: https://github.com/opencontainers/runc/blob/eb4aeed24ffbf8e2d740fafea39d91faa0ee84d0/utils_linux.go#L154

lib/box/tty.go
```diff
@@ -0,0 +1,164 @@
...
+	epoller   *console.Epoller
+	console   *console.EpollConsole
+	stdin     console.Console
+	closers   []io.Closer
+	postStart []io.Closer
+	wg        sync.WaitGroup
+	consoleC  chan error
+}
+
+func (t *tty) recvtty(process *libcontainer.Process, socket *os.File) (Err error) {
```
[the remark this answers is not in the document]
knisbet: Yea I'm a bit on the fence with the copy/paste. I did add in the trace.Wraps, but I sort of feel we're in a better position if we don't make substantial changes here unless needed.

tool/planet/etcd.go
```diff
@@ -176,11 +173,18 @@ func etcdDisable(upgradeService, stopAPIServer bool) error {
 }
 
 // etcdEnable enables a disabled etcd node
-func etcdEnable(upgradeService bool) error {
+func etcdEnable(upgradeService bool, joinToMaster string) error {
 	ctx, cancel := context.WithTimeout(context.Background(), EtcdUpgradeTimeout)
 	defer cancel()
 
 	if !upgradeService {
+		ctx, cancel := context.WithTimeout(context.Background(), 120*time.Minute)
```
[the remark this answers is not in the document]
knisbet: This is really only triggered by machine code, so it's really just a maximum value so that the join will die if it's not working.

lib/box/tty.go
knisbet: Yea, they don't have license headers in each file, so I've just removed it for now.

tool/planet/etcd.go
knisbet: Before, it appears the nodes are able to rejoin, simply by being part of the initial cluster env variable, even though they're not members of the cluster, etcd would accept them. This is just supposition, as I didn't track down the exact behaviour change. 

After going to etcd 3.4, it seems to be more strict, and not allowing this style of join. Another possibility, is that by fixing the initial cluster status to existing, which we were incorrectly setting to a new cluster previously, was also allowing the nodes to join the cluster previously as an initial cluster config. Either way, the changes prevent a node from simply starting and joining the cluster, without a member object.

tool/planet/etcd.go
knisbet: I don't think so. All data within this directory, would be the result of a previous attempt to run this phase of the upgrade. If it failed, the data is incorrect, and if it succeeded, rerunning the step should produce the same output. 

I guess the only place where the data could be useful and different, is if someone got past the etcd upgrade, and force re-run this particular phase, it would then be missing later writes. However, if the cluster is rolled back prior to this step, the data will also be cleaned up, as the rollback goes back to the previous cluster state.

tool/planet/enter.go
```diff
@@ -46,84 +45,51 @@ func enterConsole(rootfs, socketPath, cmd, user string, tty bool, stdin bool, ar
...
 
-	return s.Enter(*cfg)
+	return trace.Wrap(box.Enter(constants.RuncDataDir, cfg))
 }
 
 // stop interacts with systemctl's halt feature
-func stop(rootfs, socketPath string) error {
-	log.Infof("stop: %v", rootfs)
+func stop() error {
+	log.Infof("stop planet container")
```
a-palchikov: log.Info~~f~~

tool/planet/etcd.go
```diff
@@ -412,23 +523,100 @@ func getBaseEtcdDir(version string) string {
...
+		CAFile:    DefaultEtcdctlCAFile,
+	}
+	client, err := etcdConf.NewClient()
+	if err != nil {
+		return trace.Wrap(err)
+	}
+
+	// wait for the temporary etcd instance to complete startup
+	log.Info("Waiting for etcd initialization to complete")
+	err = waitEtcdHealthyTimeout(context.TODO(), 2*time.Minute, client)
```
a-palchikov: There's a share of arbitrary timeouts - do we want to keep that in a configuration or update on a need by need basis?

tool/planet/etcd.go
```diff
@@ -437,6 +625,32 @@ func etcdRestore(file string) error {
...
+	mapi := etcd.NewMembersAPI(client)
+
+	ticker := time.NewTicker(1 * time.Second)
+	defer ticker.Stop()
+	for {
+		select {
+		case <-ctx.Done():
+			return trace.Wrap(ctx.Err())
+		case <-ticker.C:
+			leader, _ := mapi.Leader(ctx)
```
r0mant: If the ignored argument here is an error, shouldn't we at least log it?

r0mant (approved):
Looks good with a couple of recent comments from @a-palchikov and I.

tool/planet/etcd.go
knisbet: So I'm generally of the position that because this is called by machine code, that this should simply be embedded at the call site. We don't really have a configuration system in place to allow overrides in a consistent fashion by setting some sort of background configuration file. If that were the case, then I would consider this for inclusion.

tool/planet/etcd.go
knisbet: Yea, I didn't want to log in the loop, because a number of errors are expected there. So I've updated it to capture the error, and include the error when returning due to timeout.

a-palchikov (approved):

PR #569, opened 2020-03-04
[5.5.x] add profiling endpoint and easier way to switch to debug

a-palchikov:
This PR adds support for profiling behind the joint health endpoint and a way to switch the running agent to debug mode.
Oh, and turns on the pretty flag for `planet status`.

Requires https://github.com/gravitational/satellite/pull/165.


knisbet:
So, I'm not sure if the profiler should be enabled by default. Although I think it's properly protected by mTLS because we require mTLS for everything on the healthz port, I'm not sure that leaving a profiler on/running would meet security best practices or not.

tool/planet/agent.go
```diff
@@ -306,18 +306,33 @@ func runAgent(conf *agent.Config, monitoringConf *monitoring.Config, leaderConf
...
 	return nil
 }
 
+func watchSignals(doneC chan<- struct{}) {
+	signalC := make(chan os.Signal, 2)
+	signal.Notify(signalC, os.Interrupt, syscall.SIGTERM, syscall.SIGUSR1)
+
+	for sig := range signalC {
+		if sig == syscall.SIGUSR1 {
+			switchLoggingToDebug()
```
r0mant: Should we add ability to switch it back to non-debug too?

vendor/github.com/gravitational/satellite/agent/server.go
```diff
@@ -180,15 +187,16 @@ func newRPCServer(agent *agent, caFile, certFile, keyFile string, rpcAddrs []str
...
-	// TODO: why does server need to handle both gRPC and HTTPS queries?
-	handler := grpcHandlerFunc(server, healthzHandler)
+	healthHandler := grpcHandlerFunc(server, newHealthHandler(server))
 
-	for _, addr := range rpcAddrs {
-		srv := newHTTPServer(addr, tlsConfig, handler)
+	for _, addr := range addrs {
+		handler := healthHandler
+		if addr.IP.IsLoopback() {
+			handler = grpcHandlerFunc(server, newHealthHandlerWithDebug(server))
```
knisbet: Hmm, so I'm not sure if always having the pprof endpoint running but bound only to localhost will totally address the potential for exploit. It may really be weaknesses in other areas of our implementation though, specifically, the issue with a localhost port is that even when protected by mTLS, any user on the system can connect to that port over localhost. 

And I'm not sure we're careful enough with our certificate issuance to ensure non-root/non-admin users aren't issued a cert and key that would be allowed by this endpoint, and that those users wouldn't have local/unprivileged access to the system.

tool/planet/main.go
```diff
@@ -201,6 +201,16 @@ func run() error {
 		cstatusClientKeyFile = cstatus.Flag("client-key-file", "mTLS client key file").
 					Default(ClientRPCKeyPath).OverrideDefaultFromEnvar(EnvPlanetAgentClientKeyFile).String()
 
+		cstatusDump        = app.Command("debug-status", "Dump Planet Agent debug internals")
+		cstatusDumpProfile = cstatusDump.Flag("profile", "Name of the profile to dump").Required().String()
```
bernardjkim, line 205: Providing invalid profile arg just outputs```[ERROR]: invalid profile: ...```
Maybe we could list valid profiles or add link to 
https://golang.org/pkg/runtime/pprof/#Profile

bernardjkim:
@a-palchikov could you provide some examples for using the new `debug-status` command? Just trying to understand how to utilize this cmd.

Running cmd `planet debug-status --profile=goroutine` ends up printing random characters to stdout. We could send output to a file `planet debug-status --profile=goroutine > goroutine.pprof`, but golang isn't available in planet, so we can't `go tool pprof goroutine.pprof`. Do we then copy the file into a different environment where golang is available?

tool/planet/main.go
a-palchikov: It is not really meant to be user-facing. And I'll have `goroutine` be a default. Still unsure we need the rest of the profiles.

a-palchikov:
No, we don't need a copy of go tools for that. Not sure which commit you have tried this with, but by default the `goroutine` profile should dump text stacks.

r0mant (approved):

bernardjkim (approved):

PR #570, opened 2020-03-11
Implement workaround to clean up leaking cgroups

knisbet:
This change implements a cleaner, that scans for cgroups created by
systemd-run --scope that do not have any pids assigned, indicating
that the cgroup is unused and should be cleaned up. On some systems
either due to systemd or the kernel, the scope is not being cleaned
up when the pids within the scope have completed execution, leading
to an eventual memory leak.

Kubernetes uses systemd-run --scope when creating mount points,
that may require drivers to be loaded/running in a separate context
from kubelet, which allows the above leak to occur.

https://github.com/kubernetes/kubernetes/issues/70324
https://github.com/kubernetes/kubernetes/issues/64137

Updates https://github.com/gravitational/gravity/issues/1219

### DIFF
--- a/tool/planet/agent.go
+++ b/tool/planet/agent.go
@@ -314,8 +314,6 @@ func runAgent(conf *agent.Config, monitoringConf *monitoring.Config, leaderConf 
 		}
 	}
 
-	ctx, cancel := context.WithCancel(context.Background())
-	defer cancel()
 	go runSystemdCgroupCleaner(ctx)
 
 	signalc := make(chan os.Signal, 2)
--- a/tool/planet/agent.go
+++ b/tool/planet/agent.go
@@ -314,6 +314,10 @@ func runAgent(conf *agent.Config, monitoringConf *monitoring.Config, leaderConf 
 		}
 	}
 
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+	go runSystemdCgroupCleaner(ctx)
+
 	signalc := make(chan os.Signal, 2)
 	signal.Notify(signalc, os.Interrupt, syscall.SIGTERM)
 
--- a/tool/planet/cgroup.go
+++ b/tool/planet/cgroup.go
@@ -421,7 +421,7 @@ func cleanSystemdScopes() error {
 		log.WithFields(logrus.Fields{
 			"path": procsPath,
 			"unit": unitName,
-		}).Warn("Stopped systemd scope unit with no pids.")
+		}).Info("Stopped systemd scope unit with no pids.")
 	}
 
 	return nil
--- a/tool/planet/cgroup.go
+++ b/tool/planet/cgroup.go
@@ -345,7 +345,10 @@ func runSystemdCgroupCleaner(ctx context.Context) {
 	for {
 		select {
 		case <-ticker.C:
-			cleanSystemdScopes()
+			err := cleanSystemdScopes()
+			if err != nil {
+				logrus.WithError(err).Warn("Failed to clean systemd scopes that don't contain processes")
+			}
 		case <-ctx.Done():
 			return
 		}
@@ -387,7 +390,7 @@ func cleanSystemdScopes() error {
 
 	for _, path := range paths {
 		unitName := filepath.Base(path)
-		log = log.WithFields(logrus.Fields{
+		log := log.WithFields(logrus.Fields{
 			"path": path,
 			"unit": unitName,
 		})
--- a/tool/planet/cgroup.go
+++ b/tool/planet/cgroup.go
@@ -37,12 +37,16 @@ package main
 
 import (
 	"bytes"
+	"context"
 	"fmt"
 	"io/ioutil"
+	"os"
 	"path"
+	"path/filepath"
 	"runtime"
 	"strings"
 	"text/template"
+	"time"
 
 	systemdDbus "github.com/coreos/go-systemd/dbus"
 	"github.com/davecgh/go-spew/spew"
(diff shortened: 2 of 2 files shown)


tool/planet/agent.go
```diff
@@ -314,6 +314,10 @@ func runAgent(conf *agent.Config, monitoringConf *monitoring.Config, leaderConf
 		}
 	}
 
+	ctx, cancel := context.WithCancel(context.Background())
```
r0mant: There's a context already available above - could use that one?

tool/planet/cgroup.go
```diff
@@ -320,3 +324,146 @@ func u64(n uint64) *uint64 {
...
+// https://github.com/kubernetes/kubernetes/issues/64137
+// https://github.com/gravitational/gravity/issues/1219
+//
+// Kubernetes is using systemd-run --scope when creating mounts within systemd.
+// https://github.com/gravitational/kubernetes/blob/1c045a09db662c6660562d88deff2733ca17dcf8/pkg/util/mount/mount_linux.go#L108-L131
+//
+// To clean up this leak, we want to scan for run-xxx.scope cgroups that do not have any processes, are atleast a minute
+// old, and then have systemd remove scopes that do not hold any processes / where all processes have exited.
+func runSystemdCgroupCleaner(ctx context.Context) {
+	ticker := time.NewTicker(15 * time.Minute)
```
r0mant: defer ticker.Stop()

tool/planet/cgroup.go
```diff
@@ -320,3 +324,146 @@ func u64(n uint64) *uint64 {
...
+		case <-ticker.C:
+			cleanSystemdScopes()
+		case <-ctx.Done():
+			return
+		}
+	}
+}
+
+func cleanSystemdScopes() error {
+	log := logrus.WithField(trace.Component, "clean-cgroup")
```
r0mant: Nit: The trace component name gets truncated in the logs so I'd prob. name this "cgroup-cleaner" so at least the first "cgroup" part is fully visible and can be searched for.

tool/planet/cgroup.go
```diff
@@ -320,3 +324,146 @@ func u64(n uint64) *uint64 {
...
+
+	conn, err := systemdDbus.New()
+	if err != nil {
+		return trace.Wrap(err)
+	}
+	defer conn.Close()
+
+	var paths []string
+
+	baseTime := time.Now().Add(-time.Minute)
```
r0mant: Do you think a 1 minute interval is enough? Maybe, to be on a safe side, make it like an hour, or at least 10 minutes?

tool/planet/cgroup.go
```diff
@@ -320,3 +324,146 @@ func u64(n uint64) *uint64 {
...
+
+	root := "/sys/fs/cgroup/systemd/"
+	err = filepath.Walk(root, func(path string, info os.FileInfo, err error) error {
+		if err != nil {
+			return trace.Wrap(err)
```
r0mant: trace.ConvertSystemError?

tool/planet/cgroup.go
```diff
@@ -320,3 +324,146 @@ func u64(n uint64) *uint64 {
...
+		// http://man7.org/linux/man-pages/man7/cgroups.7.html
+		// The cgroup.procs file can be read to obtain a list of the processes
+		// that are members of a cgroup.  The returned list of PIDs is not guar‐
+		// anteed to be in order.  Nor is it guaranteed to be free of dupli‐
+		// cates.  (For example, a PID may be recycled while reading from the
+		// list.)
+
+		procsPath := filepath.Join(path, "cgroup.procs")
+		pids, err := ioutil.ReadFile(procsPath)
+		if err != nil && !trace.IsNotFound(err) {
```
r0mant: I think this should be `&& !trace.IsNotFound(trace.ConvertSystemError(err))`.

tool/planet/cgroup.go
```diff
@@ -320,3 +324,146 @@ func u64(n uint64) *uint64 {
...
+
+		unitName := filepath.Base(path)
+		_, err = conn.StopUnit(unitName, "replace", nil)
+		if err != nil {
+			log.WithFields(logrus.Fields{
+				logrus.ErrorKey: err,
+				"path":          procsPath,
+				"unit":          unitName,
+			}).Warn("Failed to stop systemd unit.")
+		}
```
r0mant: Should be `continue` here, otherwise it will log "stopped systemd unit" below as well.

tool/planet/cgroup.go
```diff
@@ -320,3 +324,146 @@ func u64(n uint64) *uint64 {
...
+
+		log.WithFields(logrus.Fields{
```
r0mant: Nit: To make it a bit cleaner, I'd create a logger with these fields above and use it everywhere, e.g.

```go
logger := log.WithField("path", procsPath).WithField("unit", unitName)
...
logger.WithError(err).Warn("Failed to stop systemd unit.")
...
logger.Info("Stopped systemd scope unit with no pids.")
```

tool/planet/cgroup.go
knisbet: I was thinking the likeliness of a race here is something like a few ns, maybe a few ms if the system is busy. This should be systemd creating a scope, which still start empty, and then placing the launched process into the particular cgroup. So the window is extremely tiny, and a minute is already overkill. If the logic here is incorrect and this stops scopes it shouldn't, I don't see much difference between stopping a scope that is 1 minute old and 1 hour old to cause problems, except that 1 minute old is hopefully a lot more apparent.

I suppose the potential for a false positive, is if a process inside planet specifically creates a scope, and then doesn't use it for a few minutes, that some other process will have manually come in and cleaned it up. 

It would be possible to get the scope object from systemd first, and match specifically against Kubernetes mounts, but I was trying to keep the implementation relatively simple.

tool/planet/cgroup.go
```diff
@@ -320,3 +324,144 @@ func u64(n uint64) *uint64 {
...
+// To clean up this leak, we want to scan for run-xxx.scope cgroups that do not have any processes, are atleast a minute
+// old, and then have systemd remove scopes that do not hold any processes / where all processes have exited.
+func runSystemdCgroupCleaner(ctx context.Context) {
+	ticker := time.NewTicker(15 * time.Minute)
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-ticker.C:
+			cleanSystemdScopes()
```
a-palchikov: No way to know why the process might be failing - should log the error.

tool/planet/cgroup.go
```diff
@@ -320,3 +324,144 @@ func u64(n uint64) *uint64 {
...
+		}
+		return nil
+	})
+	if err != nil {
+		log.WithError(err).Warn("Error occurred while scanning cgroup hierarchy for unused systemd scopes.")
+	}
+
+	for _, path := range paths {
+		unitName := filepath.Base(path)
+		log = log.WithFields(logrus.Fields{
```
a-palchikov: nit: this changes the `log` from outside the loop but it probably should not - simply shadow it with `:=` so the change is only for the loop scope.

a-palchikov (approved):

PR #575, opened 2020-03-13
Agent state reconciler loop

a-palchikov:
This PR implements a reconciler loop for the planet agent that periodically (for each change in state and on general resync interval) changes the system state to match the changes in configuration and environment. This includes:
  * updating CoreDNS hosts file with the address of the new leader
  * starting control plane services on the leader and stopping them on every other master node
  * enabling/disabling election participation

Requires https://github.com/gravitational/coordinate/pull/13.
Updates https://github.com/gravitational/gravity/issues/1209.


tool/planet/agent.go
```diff
@@ -121,48 +116,59 @@ func startLeaderClient(conf *LeaderConfig, errorC chan error) (leaderClient io.C
...
 				return nil, trace.Wrap(err)
 			}
 		case false:
-			log.Info("shutting down services until election has been re-enabled")
-			// Shut down services at startup if running as master
-			if err := unitsCommand("stop"); err != nil {
-				log.Infof("failed to stop units: %v", err)
+			logger.Info("Shut down services until election has been re-enabled.")
+			if err := mon.stop(ctx); err != nil {
+				logger.WithError(err).Warn("Failed to stop control plane units.")
```
r0mant: Should this also return here? Otherwise the services will just keep running?

tool/planet/agent.go
```diff
@@ -173,34 +179,38 @@ func startLeaderClient(conf *LeaderConfig, errorC chan error) (leaderClient io.C
...
+			voterCancel()
+			voterCancel = nil
 		}
 	})
 	// modify /etc/hosts upon election of a new leader node
 	client.AddWatchCallback(conf.LeaderKey, conf.Term/3, func(key, prevVal, newLeaderIP string) {
-		if err := updateDNS(conf, hostname, newLeaderIP); err != nil {
-			log.Error(trace.DebugReport(err))
+		if err := updateDNS(conf, newLeaderIP); err != nil {
+			log.WithError(err).WithField("leader-addr", newLeaderIP).Error("Failed to update DNS.")
```
r0mant: Not really related to your change, but I wonder if the way we handle these errors (here when updating DNS and above when stopping/starting services) can be potentially causing some issues with our leader election. For example, if we failed to update it here, shouldn't we keep retrying? Otherwise this node will be talking to a wrong apiserver until the next leader change.

tool/planet/agent.go
```diff
@@ -232,27 +254,95 @@ var electedUnits = []string{
...
+
+func (r *unitMonitor) stop(ctx context.Context) error {
+	stopC := make(chan string, len(r.units))
+	// Number of successfully started stop jobs
+	var numStopped int
+	for _, unit := range r.units {
+		logger := log.WithField("unit", unit)
+		logger.Info("Shut down service.")
+		if _, err := r.conn.StopUnit(unit, "replace", stopC); err != nil {
+			log.WithError(err).WithField("unit", unit).Warn("Failed to stop unit.")
```
r0mant: You already have `logger` with "unit" field: `logger.WithError(err).Warn("")`

tool/planet/agent.go
```diff
@@ -232,27 +254,95 @@ var electedUnits = []string{
...
-	for _, unit := range electedUnits {
-		cmd := exec.Command("/bin/systemctl", command, unit)
-		log.Debugf("executing %v", cmd)
-		out, err := cmd.CombinedOutput()
-		if err != nil {
-			errors = append(errors, err)
-			// Instead of failing immediately, complete start of other units
-			log.Warningf("failed to execute %v: %s\n%v", cmd, out, trace.DebugReport(err))
+func newUnitMonitor() (*unitMonitor, error) {
+	conn, err := systemdDbus.New()
```
r0mant: So I don't know how this library works, but the potential source of new problems I see here is - could this connection somehow "go bad"? Which would effectively make this agent to be unable to stop/start units. Before, we just used systemctl directly so these kinds of issues were abstracted from us. Using cli directly is not the most pretty way but I'm just afraid this approach might reduce stability instead of improving it.

tool/planet/agent.go
```diff
@@ -232,27 +254,95 @@ var electedUnits = []string{
...
+		if _, err := r.conn.StopUnit(unit, "replace", stopC); err != nil {
+			log.WithError(err).WithField("unit", unit).Warn("Failed to stop unit.")
+			continue
+		}
+		numStopped += 1
+	}
+	if numStopped != 0 {
+		r.waitForJobs(ctx, stopC, numStopped)
+	}
+	// NB: ListUnitsByPatterns is not supported on systemd 219
```
r0mant: What's "NB"?

tool/planet/agent.go
```diff
@@ -232,27 +254,95 @@ var electedUnits = []string{
...
+	}
+	for _, unit := range failedUnits {
+		if err := r.conn.ResetFailedUnit(unit.Name); err != nil {
+			log.WithError(err).WithField("unit", unit.Name).Warn("Failed to reset failed unit.")
+		}
+	}
+	return nil
+}
+
+func (r *unitMonitor) waitForJobs(ctx context.Context, stopC <-chan string, services int) {
```
r0mant: This looks like the same could be achieved by using wait groups from stdlib.

tool/planet/agent.go
```diff
@@ -121,48 +116,59 @@ func startLeaderClient(conf *LeaderConfig, errorC chan error) (leaderClient io.C
 		}
 	}
 
+	mon, err := newUnitMonitor()
+	if err != nil {
+		return nil, trace.Wrap(err)
+	}
+	defer func() {
```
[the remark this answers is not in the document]
a-palchikov: Because it has to continue running after the function exits.

tool/planet/agent.go
a-palchikov: Which aspect of this exactly do you mean to replace with waitgroups?
`stopC` channel goes into the `go-systemd` library which will be sending on it.

tool/planet/units.go
```diff
@@ -0,0 +1,116 @@
...
+		if _, err := r.conn.StartUnit(unit, "replace", resultC); err != nil {
+			logger.WithError(err).Warn("Failed to start unit.")
+			continue
+		}
+		numStarted += 1
+	}
+	if numStarted == 0 {
+		return nil
+	}
+	r.waitForJobs(ctx, resultC, numStarted)
```
knisbet: Overall, I'm not sure I understand of waiting for the jobs resultC is? It only seems to block execution, so is the idea to not try and send too many commands to systemd is things are flapping?

tool/planet/units.go
```diff
@@ -0,0 +1,116 @@
...
+}
+
+func (r *unitMonitor) waitForJobs(ctx context.Context, resultC <-chan string, numStarted int) {
+	var done int
+	for {
+		select {
+		case result := <-resultC:
+			log.WithField("result", result).Info("Job done.")
+			done += 1
+			if done >= numStarted {
```
knisbet: I'm curious whether it's possible that the resultC could miss an event with dbus (ie it get's disconnected, or otherwise lost somehow). If that's happens, trying to trace the code, it looks like waitForJobs could block indefinitely, as the context passed afaict is the context for the overall agent service, and won't timeout until the process is requested to exit.

knisbet (commented):
Just a couple items I'm not sure about. Also I wanted to mention, I might've attempted to take a slightly different approach, which would be to re-organize this into a  desired state / control loop, that continually tries to ensure systemd is in the requested state.

tool/planet/units.go
a-palchikov: There's a chance if the systemd package or dbus daemon (I guess being the one sending the signal) does not deliver. Otherwise, how can it miss the event (with go-systemd package sending it unconditionally if provided)?

tool/planet/units.go
a-palchikov: Yes, that's the idea. Usually the notifications are pretty snappy but who knows.

tool/planet/units.go
knisbet: Yea, sorry that's sort of what I was getting at, is we're sort of relying on an external behaviour, that there isn't the possibility that it doesn't send or drops events in certain specific conditions.

tool/planet/agent.go
```diff
@@ -108,151 +104,126 @@ func startLeaderClient(conf *LeaderConfig, errorC chan error) (leaderClient io.C
...
 	return trace.Wrap(err)
 }
 
-func updateDNS(conf *LeaderConfig, hostname string, newMasterIP string) error {
-	log.Infof("Setting new leader address to %v in %v", newMasterIP, CoreDNSHosts)
-	err := writeLocalLeader(CoreDNSHosts, newMasterIP)
-	if err != nil {
-		return trace.Wrap(err)
+func updateControlPlaneUnits(localAddr string) handlerFunc {
+	return func(ctx context.Context, _, newLeaderAddr string, logger log.FieldLogger) error {
```
r0mant: Is the ignored (`_`) parameter here the old leader address? Why not use it as a logger field as well? Same below.

tool/planet/agent.go
```diff
@@ -108,151 +104,126 @@ func startLeaderClient(conf *LeaderConfig, errorC chan error) (leaderClient io.C
...
+		logger.WithField("leader-addr", newLeaderAddr).Info("New leader.")
+		if newLeaderAddr == localAddr {
+			logger.Info("Start control plane units.")
+			updateUnits(ctx, (*unitMonitor).start)
```
r0mant: `updateUnits` can return an error as far as I can see.

tool/planet/agent.go
```diff
@@ -108,151 +104,126 @@ func startLeaderClient(conf *LeaderConfig, errorC chan error) (leaderClient io.C
...
+			return nil
+		}
+		logger.Info("Shut down control plane units.")
+		updateUnits(ctx, (*unitMonitor).stop)
```
r0mant: Same here: handle an error?

tool/planet/agent.go
```diff
@@ -108,151 +104,126 @@ func startLeaderClient(conf *LeaderConfig, errorC chan error) (leaderClient io.C
...
 }
 
-var electedUnits = []string{
-	"kube-controller-manager.service",
-	"kube-scheduler.service",
-	"kube-apiserver.service",
+func updateElectionParticipation(leaderKey, localAddr string, term time.Duration, client *leader.Client) handlerFunc {
+	return func(ctx context.Context, _, enabledFlag string, logger log.FieldLogger) error {
+		var err error
+		enabled, _ := strconv.ParseBool(enabledFlag)
```
r0mant: Log the error at least?

tool/planet/agent.go
```diff
@@ -108,151 +104,126 @@ func startLeaderClient(conf *LeaderConfig, errorC chan error) (leaderClient io.C
...
+			cancel = nil
+		default:
+			if cancel != nil {
+				localDone = nil
+				cancel()
+				wg.Wait()
+			}
 		}
+		localCtx, cancel = context.WithCancel(ctx)
+		localDone = localCtx.Done()
```
r0mant: Not sure I understand how this works - `localDone` is first defined here but used in the select statement above? Also, I don't see where `localCtx` is used.

tool/planet/units.go
```diff
@@ -0,0 +1,116 @@
+package main
```
r0mant: Missing license header.

tool/planet/units.go
r0mant: Yes, (potentially) infinite loop is a bit concerning, I think there should be a some sort of timeout here. Maybe pass a `context.WithTimeout()` here.

tool/planet/units.go
```diff
@@ -0,0 +1,116 @@
...
+	}, nil
+}
+
+type unitMonitor struct {
+	conn  *systemdDbus.Conn
+	units []string
+}
+
+func (r *unitMonitor) close() {
+	r.conn.Close()
```
r0mant: Can this return an error?

tool/planet/units.go
```diff
@@ -0,0 +1,116 @@
...
+func (r *unitMonitor) start(ctx context.Context) error {
+	resultC := make(chan string, len(r.units))
+	// Number of successfully started start jobs
+	var numStarted int
+	for _, unit := range r.units {
+		logger := log.WithField("unit", unit)
+		logger.Info("Start service.")
+		if _, err := r.conn.StartUnit(unit, "replace", resultC); err != nil {
+			logger.WithError(err).Warn("Failed to start unit.")
+			continue
```
r0mant: Hmm, why do we only log this error and keep going? This way the caller will still think that everything has been started successfully, no? Can we collect all errors in trace.Aggregate, return to the caller and have it retry (I think you already have retry on error logic) so it can start units that failed to start before? Otherwise this function never actually returns an error it seems like.

tool/planet/units.go
```diff
@@ -0,0 +1,116 @@
...
+func (r *unitMonitor) stop(ctx context.Context) error {
+	resultC := make(chan string, len(r.units))
+	// Number of successfully started stop jobs
+	var numStarted int
+	for _, unit := range r.units {
+		logger := log.WithField("unit", unit)
+		logger.Info("Shut down service.")
+		if _, err := r.conn.StopUnit(unit, "replace", resultC); err != nil {
+			logger.WithError(err).Warn("Failed to stop unit.")
+			continue
```
r0mant: Same comment as above - does it make sense to return errors to the caller and have it retry?

tool/planet/units.go
a-palchikov: I'll log the error if there's one.

tool/planet/agent.go
```diff
@@ -108,151 +104,126 @@ func startLeaderClient(conf *LeaderConfig, errorC chan error) (leaderClient io.C
...
-			log.Infof("adding voter for IP %v", conf.PublicIP)
-			ctx, cancelVoter = context.WithCancel(context.TODO())
-			if err = client.AddVoter(ctx, conf.LeaderKey, conf.PublicIP, conf.Term); err != nil {
+		etcdapi := etcd.NewKeysAPI(etcdClient)
+		// Set initial value of election participation mode
+		_, err = etcdapi.Set(context.TODO(), conf.ElectionKey,
+			strconv.FormatBool(conf.ElectionEnabled),
+			&etcd.SetOptions{PrevExist: etcd.PrevNoExist})
+		if err != nil {
+			if err = convertError(err); !trace.IsAlreadyExists(err) {
```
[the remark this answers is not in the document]
a-palchikov: I don't think this is necessary - I've updated this to ignore previous value as agent might restart multiple times and this expecting to see no previous value does not make sense.

lib/reconcile/reconcile.go
```diff
@@ -0,0 +1,142 @@
...
+				logger.WithError(err).Warn("Failed to execute plan.")
+				continue
+			}
+			tickerC = nil
+			ticker.Stop()
+			cancel()
+
+		case c := <-r.stop:
+			cancel()
+			r.stop = nil
```
r0mant, line 113: Should ticker also be reset/stopped here so it doesn't fire if it's still on backoff?

r0mant (commented):
Will probably need to look through one more time.

tool/planet/agent.go
a-palchikov: Oh, missed that one - will add to the list of fields.

PR #579, opened 2020-03-18
(6.1) Upgrade serf to v0.8.5.

r0mant:
Upgrade serf to v0.8.5 to take advantage of a new `-prune` flag that allows to immediately evict a member when executing a `force-leave` command.

Note: serf v0.8.5 uses new protocol version 5 by default which older agents do not recognize so I set it to use the previous version 4. We will be able to get rid of this flag when serf agents in all supported upgrade paths understand it.

### DIFF
--- a/build.assets/makefiles/base/agent/agent.mk
+++ b/build.assets/makefiles/base/agent/agent.mk
@@ -1,6 +1,6 @@
 .PHONY: all
 
-VER := v0.8.0
+VER := v0.8.5
 REPODIR := $(GOPATH)/src/github.com/hashicorp/serf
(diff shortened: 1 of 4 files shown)


build.assets/makefiles/base/agent/agent.mk
```diff
@@ -1,6 +1,6 @@
 .PHONY: all
 
-VER := v0.8.0
+VER := v0.8.5
```
a-palchikov: I'd vote to move this to the top-level Makefile for better visibility.

a-palchikov (approved):

PR #529, opened 2019-11-16
Changes to support OpenEBS integration.

r0mant:
Minor tweaks to planet to make OpenEBS work.

PR #595, opened 2020-04-08
(6.1) Changes for openebs

r0mant:
Backport https://github.com/gravitational/planet/pull/529 to 6.1.

### DIFF
--- a/build.assets/makefiles/master/k8s-master/kube-kubelet.service
+++ b/build.assets/makefiles/master/k8s-master/kube-kubelet.service
@@ -9,6 +9,7 @@ EnvironmentFile=/run/dns.env
 EnvironmentFile=-/run/kubernetes-reserved.env
 ExecStartPre=/bin/systemctl is-active docker.service
 ExecStart=/usr/bin/kubelet \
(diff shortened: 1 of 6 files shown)


build.assets/makefiles/master/k8s-master/kube-kubelet.service
```diff
@@ -9,6 +9,7 @@ EnvironmentFile=/run/dns.env
 EnvironmentFile=-/run/kubernetes-reserved.env
 ExecStartPre=/bin/systemctl is-active docker.service
 ExecStart=/usr/bin/kubelet \
+        --root-dir=/var/lib/gravity/planet/kubelet \
```
a-palchikov, line 12: I'd have to change this to `/var/lib/kubelet` (so it's the same as the kubelet's default) when I port the SELinux changes to 6.1.x. The reason I kept the `root-dir` set to the default (and it should be irrelevant for this PR really, as it is just configuration in gravity) is that kubelet hardcodes the default root directory in some paths resulting in issues (well, inconsistency in path handling) as we advertise planet file contexts to SELinux.

a-palchikov (approved):

PR #596, opened 2020-04-09
Tune image pulls within kubelet for large images

knisbet:
Updates https://github.com/gravitational/gravity/issues/1167

Tune kubelet parameters to allows parallel image pulls and increase the progress deadline from 1m to 15m. 

Increasing the deadilne should be relatively safe, but I'm not sure how risky changing the serialize-image-pulls setting is. Based on some descriptions, aufs apparently had some corruption problems with parallel pulls, which isn't present when using overlay2. On my test cluster, I didn't observe any issues in limited testing, and confirmed that image pulls are not blocked while a large image is still downloading.

### DIFF
--- a/build.assets/makefiles/master/k8s-master/kube-kubelet.service
+++ b/build.assets/makefiles/master/k8s-master/kube-kubelet.service
@@ -22,7 +22,6 @@ ExecStart=/usr/bin/kubelet \
         --system-reserved=${KUBE_SYSTEM_RESERVED} \
         --cgroup-root=${KUBE_CGROUP_ROOT} \
         --config=/etc/kubernetes/kubelet.yaml \
-        --serialize-image-pulls=false \
         --image-pull-progress-deadline=15m \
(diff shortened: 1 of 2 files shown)


build.assets/makefiles/master/k8s-master/kube-kubelet.service
```diff
@@ -22,6 +22,8 @@ ExecStart=/usr/bin/kubelet \
         --system-reserved=${KUBE_SYSTEM_RESERVED} \
         --cgroup-root=${KUBE_CGROUP_ROOT} \
         --config=/etc/kubernetes/kubelet.yaml \
+        --serialize-image-pulls=false \
```
a-palchikov, line 25: At least this should be set in kubelet configuration file [here](https://github.com/gravitational/planet/blob/kevin/master/1167-image-pull-deadline/tool/planet/constants.go#L471) (as the CLI option is marked as deprecated):
```go
// KubeletConfig specifies the default kubelet configuration
var KubeletConfig = kubeletconfig.KubeletConfiguration{
	// ...
	SerializeImagePulls: utils.BoolPtr(false),
}
```

a-palchikov (approved):

tool/planet/constants.go
```diff
@@ -513,6 +513,9 @@ var KubeletConfig = kubeletconfig.KubeletConfiguration{
 	Authorization: kubeletconfig.KubeletAuthorization{
 		Mode: kubeletconfig.KubeletAuthorizationModeWebhook,
 	},
+	// Allow parallel image pulls to avoid head of line blocking
+	// https://github.com/gravitational/gravity/issues/1167
+	SerializeImagePulls: utils.BoolPtr(false),
```
r0mant, line 518: I hope kubelet has some sort of limit/backoff on this internally so we don't destroy nodes with subpar i/o performance.

r0mant (approved):

PR #597, opened 2020-04-09
Record leader elected event

bernardjkim:
Planet agents will now record a new event whenever they are elected as leader.

Also includes patch to nethealth-checker https://github.com/gravitational/satellite/pull/183

Requires https://github.com/gravitational/satellite/pull/167


tool/planet/agent.go
```diff
@@ -148,6 +148,26 @@ func startLeaderClient(conf *LeaderConfig, errorC chan error) (leaderClient io.C
...
 	})
 
+	// Add a callback to watch for changes to the leader key.
+	// If this node becomes the leader, record a LeaderElected event to the
+	// local timeline.
+	client.AddWatchCallback(conf.LeaderKey, conf.Term/3, func(key, prevVal, newVal string) {
+		// recordEventTimeout specifies the max timeout to record an event
+		const recordEventTimeout = 10 * time.Second
+
+		if newVal != conf.PublicIP {
```
r0mant, line 158: I wonder if that might be helpful for every node to record leader change so these events can be correlated.

tool/planet/agent.go
```diff
@@ -148,6 +148,26 @@ func startLeaderClient(conf *LeaderConfig, errorC chan error) (leaderClient io.C
...
+			return
+		}
+
+		ctx, cancel := context.WithTimeout(context.Background(), recordEventTimeout)
+		defer cancel()
+
+		agent.RecordLocalEvents(ctx, []*pb.TimelineEvent{
+			pb.NewLeaderElected(agent.GetConfig().Clock.Now(),
```
r0mant, line 166: How about recording the old leader IP as well?

tool/planet/agent.go
bernardjkim: I think adding an entry for each node will end up creating too much noise and will make it more difficult to read through the history. Also planet is already logging the new leader elected on all nodes.

r0mant: Yeah, that's fair and I too thought it might be a bit noisy. My reasoning was to be able to tell if any node for some reason missed the new leader election but I guess this may be trying to make the timeline more into logs so we gotta keep the balance.

The information about previous leader would certainly be helpful though, so I would either add a counterpart "leadership lost" event or augment the current event with information about the previous leader.

tool/planet/agent.go
bernardjkim: https://github.com/gravitational/planet/pull/597/commits/116acba092c17e97983c1b12ddf7a4b9de347c09

r0mant (approved):

PR #604, opened 2020-04-21
Override systemd defaults and set unlimitted pids for planet container

knisbet:
Updates https://github.com/gravitational/gravity/issues/1399

### DIFF
--- a/lib/box/srv.go
+++ b/lib/box/srv.go
@@ -388,6 +388,7 @@ func getLibcontainerConfig(containerID, rootfs string, cfg Config) (*configs.Con
 				AllowedDevices:   configs.DefaultAllowedDevices,
 				MemorySwappiness: nil, // nil means "machine-default" and that's what we need because we don't care
 				CpuShares:        2,   // set planet to minimum cpu shares relative to host services
+				PidsLimit:        -1,  // override systemd defaults and set planet scope to unlimitted pids
 			},
 		},
 		Devices:  append(configs.DefaultAutoCreatedDevices, append(loopDevices, disks...)...),
--- a/lib/box/srv.go
+++ b/lib/box/srv.go
@@ -388,7 +388,7 @@ func getLibcontainerConfig(containerID, rootfs string, cfg Config) (*configs.Con
 				AllowedDevices:   configs.DefaultAllowedDevices,
 				MemorySwappiness: nil, // nil means "machine-default" and that's what we need because we don't care
 				CpuShares:        2,   // set planet to minimum cpu shares relative to host services
-				PidsLimit:        -1,  // override systemd defaults and set planet scope to unlimitted pids
+				PidsLimit:        -1,  // override systemd defaults and set planet scope to unlimited pids
 			},
 		},
 		Devices:  append(configs.DefaultAutoCreatedDevices, append(loopDevices, disks...)...),


lib/box/srv.go
```diff
@@ -388,6 +388,7 @@ func getLibcontainerConfig(containerID, rootfs string, cfg Config) (*configs.Con
 				AllowedDevices:   configs.DefaultAllowedDevices,
 				MemorySwappiness: nil, // nil means "machine-default" and that's what we need because we don't care
 				CpuShares:        2,   // set planet to minimum cpu shares relative to host services
+				PidsLimit:        -1,  // override systemd defaults and set planet scope to unlimitted pids
```
r0mant, line 391: unlimited

r0mant (approved):

a-palchikov:
According to libcontainer code (and comments), values <= 0 set no limits meaning the absence of `PidsLimit` was also setting the limit to `max`. Is there a catch?

knisbet:
@a-palchikov Possibly, but currently the parent of the scope created for planet is system.slice which is managed by the hosts systemd. So we're basically inheriting the cgroup limits that have been assigned for all system services. I think it is possible this get's configured too small, but all system services would be affected, not just planet. I wouldn't expect systemd as a default to have an arbitrarily low value for system.slice, and an admin could probably configure the setting to low.

We could look at moving planet into machine.slice (systemd's parent slice meant for VMs and containers), but last time I tried to do this I wasted a bunch of time with IIRC buggy behaviour. 

So to me it seems to be worth the risk to inherit system.slice pids.max in the planet scope, simply because all systemd services will be impacted if the max is set too low.

We could also consider a satellite check for not having enough free pids within a cgroup or a miss-configuration of system.slice.

Here is more of the structure from the system investigated at customer A:
```
/sys/fs/cgroup/pids/system.slice/pids.max:max
/sys/fs/cgroup/pids/system.slice/pids.current:752
/sys/fs/cgroup/pids/system.slice/-planet-0e6a5905-3c44-4af3-a8b4-5e275c5a6630.scope/pids.max:512
/sys/fs/cgroup/pids/system.slice/-planet-0e6a5905-3c44-4af3-a8b4-5e275c5a6630.scope/pids.current:317
/sys/fs/cgroup/pids/system.slice/-planet-0e6a5905-3c44-4af3-a8b4-5e275c5a6630.scope/system.slice/pids.max:max
/sys/fs/cgroup/pids/system.slice/-planet-0e6a5905-3c44-4af3-a8b4-5e275c5a6630.scope/system.slice/pids.current:315
```

a-palchikov:
@knisbet I was just curious what I was getting at is that the PR does not seem to change anything (i.e. the `PidsLimit` would have been set to `max` if it was left out).

knisbet:
@a-palchikov I don't seem to interpret the runc code as using max when PidsLimit isn't set.
https://github.com/opencontainers/runc/blob/db3159c9d9d3b65d2d27ff33f722eb9797f20e52/libcontainer/cgroups/fs/pids.go#L33-L48

If the PidsLimit is 0, it skips setting the pids.max within the cgroup (line 34). Only if the limit is != 0, does it go on to set the limit to max if below 0, or the specific value if above 0. Since we're initializing the struct, unless I'm missing it getting set somewhere else, PidsLimit initializes to 0, and doesn't get written to the cgroup as "max".

a-palchikov:
Oh, yes - you're correct. My bad.

PR #609, opened 2020-04-23
Specify tags for planet/os and planet/buildbox

bernardjkim:
**Rationale**
- This change fixes potential race conditions in the build process.
- This change could also help improve build speeds. We can cache multiple versions of `planet/os` and `planet/buildbox` instead of caching only the latest build.

**Bug Description**

Ran into an issue while publishing planet artifacts through jenkins. I think this issue is due to a race condition in the build process. 

We have two intermediate images `planet/os` and `planet/base` that are built during the planet build process. `planet/base` uses `planet/os` as its base image. The problem occurs when multiple builds are running concurrently on jenkins. `planet/os` is built without a specified tag, so `planet/base` just looks for `planet/os:latest` image. If there are other jenkins jobs running, the `planet/os:latest` image can get overwritten by another build. This lead to some confusion with iscsi services being enabled in the final planet image when iscsi services were supposed to be masked in the `planet/os` image.

**Note**

Maybe not an immediate concern, but this change will allow multiple different versions of `planet/os` and `planet/buildbox` to be cached. We may want to implement some sort of clean up step to free up disk.


knisbet (approved):

r0mant (approved):

a-palchikov:
In CI environment, the image can be removed after a successful (or even unsuccessful) build.

PR #616, opened 2020-04-26
[6.1.x] agent reconciler loop

a-palchikov:
Backport of https://github.com/gravitational/planet/pull/575


tool/planet/reconciler.go
```diff
@@ -0,0 +1,351 @@
...
+	if r.leaderAddr == "" {
+		return steps, nil
+	}
+	if err := verifyLeaderAddr(r.leaderAddr); err != nil {
+		logger.WithError(err).Warn("Failed to check local leader DNS configuration, will update configuration.")
+		steps = append(steps, updateDNS{leaderAddr: r.leaderAddr})
+	}
+	state := activeStateActive
+	units := serviceUnits
+	if r.localAddr != r.leaderAddr {
```
r0mant, line 47: So on the leader node, the reconciler will make sure that all services (incl. docker, kubelet, etc.) are active. But on a regular node, it only makes sure that control plane components are inactive. Is this intended? Do we need to include non control plane components at all in the reconciler?

lib/reconcile/reconcile.go
```diff
@@ -0,0 +1,200 @@
...
+func (r *Reconciler) executorLoop(ctx context.Context, plan Plan, wg *sync.WaitGroup) {
+	defer wg.Done()
+	timeout := r.config.Timeout
+	if r.executePlan(ctx, plan) == nil {
+		timeout = r.config.ResyncTimeout
+	}
+	ticker := r.config.clock.NewTicker(timeout)
+	defer ticker.Stop()
+	tickerC := ticker.Chan()
+	for {
```
r0mant, line 145: So this seems to be executing the plan forever, not just until success? Is that on purpose? I wonder if there should be some way to pause it (other than stopping planet-agent) when you need to make some changes "out-of-band" to troubleshoot something for example.

a-palchikov: The purpose of the loop is bringing the system to the desired state hence it is running forever, albeit with longer intervals with no changes to the plan.
I'm not sure about out-of-band changes short of stopping the agent - do kubernetes components (as an example) have an ability to stop their loops for out-of-band troubleshooting?

tool/planet/reconciler.go
a-palchikov: I've revised the part as it was wrong in deciding which services to stop/start.

lib/reconcile/reconcile.go
```diff
@@ -0,0 +1,200 @@
...
+		select {
+		case <-tickerC:
+			if r.executePlan(ctx, plan) != nil {
+				if timeout != r.config.Timeout {
```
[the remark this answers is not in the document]
a-palchikov: I felt like 10s would be too frequent for the resync interval.

a-palchikov:
@r0mant @knisbet Guys if you could take another look and shared your comments - I'd appreciate it! Open issues as far as I can tell:
### separate reconciler loop timeouts
I don't feel easy removing either as I think the loop should not be constantly checking the state after a successful run, but I'm open for constructive improvements or change requests.

### troubleshooting and the loop
I don't see a particular need to have a separate knob to disable the loop while troubleshooting - stopping the agent should be enough to prevent any services from getting restarted if this is required during troubleshooting. If you have other concerns that become an issue with the loop active - please do share them.

tool/planet/units.go
```diff
@@ -0,0 +1,108 @@
...
+
+func (r *unitMonitor) filterUnits(states ...serviceActiveState) ([]systemdDbus.UnitStatus, error) {
+	filter := make([]string, 0, len(states))
+	for _, state := range states {
+		filter = append(filter, string(state))
+	}
+	return r.conn.ListUnitsFiltered(filter)
+}
+
+// status validates whether the status of the specified service equals targetStatus
```
bernardjkim, line 76: Is `status` the correct function name here? It looks like this function is checking if the service exists in the list of units, rather than validating the status of the service.

tool/planet/reconciler.go
```diff
@@ -0,0 +1,425 @@
...
+	}
+}
+
+// Name returns the name of this plan.
+// Implements reconciler.Plan
+func (r *agentPlan) Name() string {
+	return "agent state reconciler"
+}
+
+// Creates a new plan for agent operation.
```
bernardjkim, line 56: nit: Creates -> Create creates

tool/planet/agent.go
```diff
@@ -128,151 +122,86 @@ func startLeaderClient(conf *LeaderConfig, agent agent.Agent, errorC chan error)
...
 		}
-	})
-	// modify /etc/hosts upon election of a new leader node
-	client.AddWatchCallback(conf.LeaderKey, conf.Term/3, func(key, prevVal, newLeaderIP string) {
-		if err := updateDNS(conf, hostname, newLeaderIP); err != nil {
-			log.Error(trace.DebugReport(err))
+		logger.WithField("leader-addr", newLeaderAddr).Info("New leader.")
+		plan := state.withLeaderAddr(newLeaderAddr)
+		reconciler.Reset(ctx, plan)
+		if newLeaderAddr != plan.localAddr {
```
bernardjkim, line 171: I think we want `if newLeaderAddr == plan.localAddr` here. We only want the new leader that was elected to record a new event.

PR #628, opened 2020-04-28
Pull OS image for every build

UnderGreen:
Found out we are building planet with the old Debian 9.1 on the Jenkins machine.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -34,6 +34,7 @@ BUILD_ASSETS := $(PWD)/build/assets
 BUILDDIR ?= $(PWD)/build
 BUILDDIR := $(shell realpath $(BUILDDIR))
 OUTPUTDIR := $(BUILDDIR)/planet
+DOCKER_BUILD_OPTS ?= "--pull"
 
 KUBE_VER ?= v1.13.12
 SECCOMP_VER ?= 2.3.1-2.1+deb9u1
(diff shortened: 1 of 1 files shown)


knisbet (approved):

Makefile
```diff
@@ -157,7 +157,7 @@ enter:
 .PHONY: os
 os:
 	@echo -e "\n---> Making Planet/OS (Debian) Docker image...\n"
-	$(MAKE) -e BUILDIMAGE=$(PLANET_OS_IMAGE) DOCKERFILE=os.dockerfile make-docker-image
+	$(MAKE) -e BUILDIMAGE=$(PLANET_OS_IMAGE) DOCKERFILE=os.dockerfile EXTRA_ARGS="--pull" make-docker-image
```
a-palchikov, line 160: Can we make it overridable from command line/environment? I tend to avoid frequent image updates during development as they can take a lot of time.

UnderGreen: @a-palchikov done!

wadells (changes requested):
I'd *strongly* suggest that we pin to a known & stationary image in os.dockerfile.  Use a hash if we can't trust the tags to be be constant.  `stretch-backports` is based on `stretch` which is not a stable tag.  It changes all the time.

a-palchikov:
@wadells put it very well. The non-determinism argument is extremely strong and pulling just for getting the latest updates is not always the best choice. I agree if we find a way to pin the specific version of the upstream image it would be the ideal solution.

UnderGreen:
I agree with both of you about pinning an image. But I believe we will end up sometimes in future some of the users will report security issues with planet images because no one touched pinned base image in 2 years.

PR #650, opened 2020-05-26
[5.5.x] bump etcd + cni plugins

knisbet:


build.assets/makefiles/base/network/network.mk
```diff
@@ -15,7 +15,7 @@ all: $(BINARIES) network.mk
 
 # Setup CNI and include flannel as a plugin
 	mkdir -p $(ROOTFS)/etc/cni/net.d/ $(ROOTFS)/opt/cni/bin
-	curl -L --retry 5 https://github.com/containernetworking/plugins/releases/download/v0.7.5/cni-plugins-amd64-v0.7.5.tgz \
+	curl -L --retry 5 https://github.com/containernetworking/plugins/releases/download/v0.8.6/cni-plugins-linux-amd64-v0.8.6.tgz \
```
a-palchikov, line 18: nit: would make sense to also move the version to the top-level Makefile

a-palchikov (approved):

r0mant (approved):

build.assets/makefiles/base/network/network.mk
knisbet: I believe it already was in later releases but not backported. I see no reason to make that change as part of a version bump.

PR #661, opened 2020-06-01
[6.1.x] Update buildbox go version

bernardjkim:
### Description
Bump buildbox go version to 1.14. Latest satellite patch uses `time.Duration.Millisecond()` which is undefined in go 1.12.9.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -50,7 +50,7 @@ ETCD_VER := v2.3.8 v3.3.4 v3.3.9 v3.3.11 v3.3.12 v3.3.22
 # This is the version of etcd we should upgrade to (from the version list)
 ETCD_LATEST_VER := v3.3.22
 
-BUILDBOX_GO_VER ?= 1.12.9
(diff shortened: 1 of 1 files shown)


a-palchikov (approved):

knisbet:
I'm a bit on the fence on this one, because I just happened to be reading yesterday about kubernetes upstream considerations on doing golang version changes within a release. 

https://groups.google.com/d/msg/kubernetes-dev/tOe2UFB_weQ/UzLP_HJUCQAJ

So it seems the risk is, that we pick up some unexpected stdlib change between releases, that doesn't stand out immediately.

r0mant (commented):
I think I'm with Kevin on this one. I would try to avoid an additional risk in LTS versions. As for the satellite calling undefined method - doesn't satellite have its own version/6.1.x branch?

bernardjkim:
@knisbet @r0mant 
Ah ok. I'll just removed the use of `duration.Millisecond()` for now.

PR #672, opened 2020-06-08
[5.5.x] Update disk check

bernardjkim:
### Description
This PR bumps satellite to 5.5.18 and now provides two separate low and high watermark values to the storage check. The values are set with `low-watermark` and `high-watermark` flags. The defaults are set at `PLANET_LOW_WATERMARK` and `PLANET_HIGH_WATERMARK`.

### Linked tickets and PRs
* Requires https://github.com/gravitational/satellite/pull/225

### DIFF
--- a/lib/monitoring/checkers.go
+++ b/lib/monitoring/checkers.go
@@ -70,7 +70,7 @@ type Config struct {
 	NodeName string
 	// LowWatermark is the disk usage warning limit percentage of monitored directories
 	LowWatermark uint
-	// HighWatermark is the disk usage critical limit percentage of monitored directores
+	// HighWatermark is the disk usage critical limit percentage of monitored directories
 	HighWatermark uint
 	// HTTPTimeout specifies the HTTP timeout for checks
 	HTTPTimeout time.Duration
--- a/lib/monitoring/checkers.go
+++ b/lib/monitoring/checkers.go
@@ -68,10 +68,10 @@ type Config struct {
 	CloudProvider string
 	// NodeName is the kubernetes name of this node
 	NodeName string
-	// WatermarkWarning is the usage limit percentage of monitored directories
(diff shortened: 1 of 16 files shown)


lib/monitoring/checkers.go
```diff
@@ -68,8 +68,12 @@ type Config struct {
 	CloudProvider string
 	// NodeName is the kubernetes name of this node
 	NodeName string
-	// HighWatermark is the usage limit percentage of monitored directories and devicemapper
+	// HighWatermark is the usage limit percentage of devicemapper
```
a-palchikov, line 71: We don't have devicemapper anymore. I would also mention disk explicitly as otherwise watermark can be about anything - i.e. RAM.
Also, what is the purpose of `HighWatermark` now - is it obsolete with the introduction of these two new attributes?

lib/monitoring/checkers.go
```diff
@@ -68,8 +68,12 @@ type Config struct {
...
 	HighWatermark uint
+	// WatermarkWarning is the usage limit percentage of monitored directories
+	WatermarkWarning uint
```
r0mant, line 74: Yeah, I agree with Dima, it's a bit confusing. Why not just reuse HighWatermark parameter, and add it's LowWatermark counterpart? TBH I'm not a big fan of "encoding" severities in the field names (i.e. XXXWarn/XXXCritical).

lib/monitoring/checkers.go
bernardjkim: It was used for the devicemapper check. I'll go ahead and remove the devicemapper check.
Will rename warning/critical -> low/high.

bernardjkim: @a-palchikov 9b8e741
Okay to remove this check?

r0mant (approved):

PR #684, opened 2020-06-16
Add apiserver flags for token volume projection

r0mant:
Forward-port https://github.com/gravitational/planet/pull/598 to master.

### DIFF
--- a/build.assets/makefiles/master/k8s-master/kube-apiserver.service
+++ b/build.assets/makefiles/master/k8s-master/kube-apiserver.service
@@ -16,6 +16,8 @@ ExecStart=/usr/bin/kube-apiserver \
         --insecure-port=0 \
         --service-account-key-file=/var/state/apiserver.key \
         --service-account-lookup=true \
+        --service-account-signing-key-file=/var/state/apiserver.key \
+        --service-account-issuer=api \
         --enable-admission-plugins=PodSecurityPolicy,NodeRestriction,AlwaysPullImages \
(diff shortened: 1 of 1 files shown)


build.assets/makefiles/master/k8s-master/kube-apiserver.service
```diff
@@ -16,6 +16,8 @@ ExecStart=/usr/bin/kube-apiserver \
         --insecure-port=0 \
         --service-account-key-file=/var/state/apiserver.key \
         --service-account-lookup=true \
+        --service-account-signing-key-file=/var/state/apiserver.key \
+        --service-account-issuer=api \
```
knisbet, line 20: It's not clear to me where the value "api" comes from, and is intended. Also, it seems that this is expected to be a URL to support some sort of Issuer Discovery as per: https://kubernetes.io/docs/reference/command-line-tools-reference/kube-apiserver/

a-palchikov: Also relevant: https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#service-account-token-volume-projection

r0mant: It can be either a string or a URI. It's just a value that will be a part of "iss" claim in the issued token, e.g. see https://jpweber.io/blog/a-look-at-tokenrequest-api/.

PR #685, opened 2020-06-16
[6.1.x] reset any control plane units that failed to stop

a-palchikov:
This PR attempts to mitigate the situation when one of the control plane units that are active on the leader, need to shut down (i.e. during fail-over or when the elections are explicitly paused) and (one of the) units fail to stop properly entering failed state which degrades cluster state as a result.

Updates https://github.com/gravitational/gravity/issues/1209.

### DIFF
--- a/tool/planet/agent.go
+++ b/tool/planet/agent.go
@@ -139,12 +139,12 @@ func startLeaderClient(conf *LeaderConfig, agent agent.Agent, errorC chan error)
 		log.Infof("new leader: %v", newVal)
 		if newVal == conf.PublicIP {
 			if err := startUnits(context.TODO()); err != nil {
-				log.Warnf("Failed to start units: %v", err)
+				log.WithError(err).Warn("Failed to start units.")
 			}
 			return
 		}
 		if err := stopUnits(context.TODO()); err != nil {
-			log.Warnf("Failed to stop units: %v.", err)
+			log.WithError(err).Warn("Failed to stop units:.")
 		}
 	})
 
@@ -264,7 +264,7 @@ func startUnits(ctx context.Context) error {
 		if err != nil {
 			errors = append(errors, err)
 			// Instead of failing immediately, complete start of other units
-			logger.WithError(err).Warnf("Failed to start unit.")
+			logger.WithError(err).Warn("Failed to start unit.")
 		}
 	}
 	return trace.NewAggregate(errors...)
@@ -278,15 +278,16 @@ func stopUnits(ctx context.Context) error {
 		err := systemctlCmd(ctx, "stop", unit)
 		if err != nil {
 			errors = append(errors, err)
-			// Instead of failing immediately, complete start of other units
-			logger.WithError(err).Warnf("Failed to stop unit: %s.")
+			// Instead of failing immediately, complete stop of other units
+			logger.WithError(err).Warn("Failed to stop unit.")
 		}
 		// Even if 'systemctl stop' did not fail, the service could have failed stopping
-		// even though 'stop' is blocking, it does not return an error upon service failing
+		// even though 'stop' is blocking, it does not return an error upon service failing.
+		// See github.com/gravitational/gravity/issues/1209 for more details
 		if err := systemctlCmd(ctx, "is-failed", unit); err == nil {
 			logger.Info("Reset failed unit.")
 			if err := systemctlCmd(ctx, "reset-failed", unit); err != nil {
-				logger.Warnf("Failed to reset failed unit: %v.", err)
+				logger.WithError(err).Warn("Failed to reset failed unit.")
 			}
 		}
 	}
--- a/tool/planet/agent.go
+++ b/tool/planet/agent.go
@@ -144,7 +144,7 @@ func startLeaderClient(conf *LeaderConfig, agent agent.Agent, errorC chan error)
 			return
 		}
 		if err := stopUnits(context.TODO()); err != nil {
-			log.WithError(err).Warn("Failed to stop units:.")
+			log.WithError(err).Warn("Failed to stop units.")
 		}
 	})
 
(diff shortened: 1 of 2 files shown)


tool/planet/agent.go
```diff
@@ -139,13 +138,13 @@ func startLeaderClient(conf *LeaderConfig, agent agent.Agent, errorC chan error)
 	client.AddWatchCallback(conf.LeaderKey, conf.Term/3, func(key, prevVal, newVal string) {
 		log.Infof("new leader: %v", newVal)
 		if newVal == conf.PublicIP {
-			if err := unitsCommand("start"); err != nil {
-				log.Infof("failed to start units: %v", err)
+			if err := startUnits(context.TODO()); err != nil {
+				log.Warnf("Failed to start units: %v", err)
```
r0mant, line 142: log.WithError?

tool/planet/agent.go
```diff
@@ -139,13 +138,13 @@ func startLeaderClient(conf *LeaderConfig, agent agent.Agent, errorC chan error)
...
 			}
 			return
 		}
-		if err := unitsCommand("stop"); err != nil {
-			log.Infof("failed to stop units: %v", err)
+		if err := stopUnits(context.TODO()); err != nil {
+			log.Warnf("Failed to stop units: %v.", err)
```
r0mant, line 147: log.WithError? There're also a few places below with the same (unless it's intended here).

tool/planet/agent.go
```diff
@@ -250,23 +249,45 @@ func updateDNS(conf *LeaderConfig, hostname string, newMasterIP string) error {
...
-		log.Debugf("executing %v", cmd)
-		out, err := cmd.CombinedOutput()
+	for _, unit := range controlPlaneUnits {
+		logger := log.WithField("unit", unit)
+		err := systemctlCmd(ctx, "start", unit)
 		if err != nil {
 			errors = append(errors, err)
 			// Instead of failing immediately, complete start of other units
-			log.Warningf("failed to execute %v: %s\n%v", cmd, out, trace.DebugReport(err))
+			logger.WithError(err).Warnf("Failed to start unit.")
```
r0mant, line 267: Nit: Warn()

tool/planet/agent.go
```diff
@@ -250,23 +249,45 @@ func updateDNS(conf *LeaderConfig, hostname string, newMasterIP string) error {
...
+	var errors []error
+	for _, unit := range controlPlaneUnits {
+		logger := log.WithField("unit", unit)
+		err := systemctlCmd(ctx, "stop", unit)
+		if err != nil {
+			errors = append(errors, err)
+			// Instead of failing immediately, complete start of other units
+			logger.WithError(err).Warnf("Failed to stop unit: %s.")
+		}
+		// Even if 'systemctl stop' did not fail, the service could have failed stopping
```
r0mant, line 284: Link relevant Github issue?

tool/planet/agent.go
```diff
@@ -250,23 +249,45 @@ func updateDNS(conf *LeaderConfig, hostname string, newMasterIP string) error {
...
+		// even though 'stop' is blocking, it does not return an error upon service failing
+		if err := systemctlCmd(ctx, "is-failed", unit); err == nil {
+			logger.Info("Reset failed unit.")
+			if err := systemctlCmd(ctx, "reset-failed", unit); err != nil {
+				logger.Warnf("Failed to reset failed unit: %v.", err)
```
r0mant, line 289: WithError

tool/planet/agent.go
```diff
@@ -250,23 +249,45 @@ func updateDNS(conf *LeaderConfig, hostname string, newMasterIP string) error {
...
+
+func stopUnits(ctx context.Context) error {
+	log.Debug("Stop control plane units.")
+	var errors []error
+	for _, unit := range controlPlaneUnits {
+		logger := log.WithField("unit", unit)
+		err := systemctlCmd(ctx, "stop", unit)
+		if err != nil {
+			errors = append(errors, err)
+			// Instead of failing immediately, complete start of other units
```
r0mant, line 281: Nit: start -> stop

r0mant (approved):
Let's backport to 5.5 before porting to other branches.

tool/planet/agent.go
```diff
@@ -250,23 +249,46 @@ func updateDNS(conf *LeaderConfig, hostname string, newMasterIP string) error {
...
+	}
+	return trace.NewAggregate(errors...)
+}
+
+func stopUnits(ctx context.Context) error {
+	log.Debug("Stop control plane units.")
+	var errors []error
+	for _, unit := range controlPlaneUnits {
+		logger := log.WithField("unit", unit)
+		err := systemctlCmd(ctx, "stop", unit)
```
knisbet, line 278: Thinking about this for a moment, I wonder with the stop commands blocking, whether it could take a long time for the apiserver to even begin shutting down here, blocked on the controller and scheduler shutdowns. 

I'm also curious whether this could be causing us any sanity issues, since it's possible for the startup on a new node after election change can race the shutdown of the old node (we tell the processes there is only a single instance running, although I suspect locks are taken regardless). It might make sense to move this to an errgroup, with each stop running it's own goroutine. On the other hand I'm not sure I have enough faith in systemd to accept rapid fire commands like that.

a-palchikov: I don't think running the commands in parallel is a good idea unless we can retry the stop for apiserver since it's in the dependency list for both scheduler and controller manager services or just ensure they're both stopped before stopping the apiserver which the old code is doing anyways.
One thing this is not doing is ensuring that the services really did shut down - if they had not, i.e. systemctl came back with an error from dbus (we've seen issues with the startup), the services might not have shut down. I'm not sure how much to invest into that as opposed to [having the reconciler loop](https://github.com/gravitational/planet/pull/616) which should fix that as well. I'd be considering this as a precursor to the reconciler loop that applies the relevant patch in the short term.
Regarding the lock between the new and the old node - yes, I suppose this could race and end up unexpectedly which I believe could still be fixed with a proper reconciler loop.

PR #688, opened 2020-06-19
Reset any control plane units that failed to stop.

a-palchikov:
Backport of https://github.com/gravitational/planet/pull/685.

Updates github.com/gravitational/gravity/issues/1209.

### DIFF
--- a/tool/planet/agent.go
+++ b/tool/planet/agent.go
@@ -23,7 +23,6 @@ import (
 	"io"
 	"io/ioutil"
 	"os"
-	"os/exec"
 	"os/signal"
 	"strconv"
 	"syscall"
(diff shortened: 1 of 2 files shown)


r0mant (approved):

tool/planet/agent.go
```diff
@@ -250,23 +249,46 @@ func updateDNS(conf *LeaderConfig, hostname string, newMasterIP string) error {
...
+		err := systemctlCmd(ctx, "stop", unit)
+		if err != nil {
+			errors = append(errors, err)
+			// Instead of failing immediately, complete stop of other units
+			logger.WithError(err).Warn("Failed to stop unit.")
+		}
+		// Even if 'systemctl stop' did not fail, the service could have failed stopping
+		// even though 'stop' is blocking, it does not return an error upon service failing.
+		// See github.com/gravitational/gravity/issues/1209 for more details
+		if err := systemctlCmd(ctx, "is-failed", unit); err == nil {
```
knisbet, line 287: Sorry, I just realized something I missed on the non-backport PR. Since we pass no-block to the underlying systemctl command, will a unit enter the failed state before we check the is-failed status, since presumably the unit is allowed some amount of time to shutdown?

a-palchikov: We don't pass `no-block` to `systemctl` here - only etcd-related functionality does - so it should be good given the possibility that dbus command could still fail which we'll address with reconciler loop changes.

knisbet: Ah my bad, I was looking at the wrong function signature when that stood out to me.

PR #697, opened 2020-07-03
[6.1.x] service CIDR update

a-palchikov:
This PR introduces changes necessary to support service CIDR updates.

### API server service
Planet agent is responsible for reconciling the API server service - it does not recreate it but rather verifies that the existing service has a valid cluster IP and if not - deletes the service, so the API server can recreate it.

### DNS services
Additionally, the agent will now use the correct DNS service address as during the ClusterConfiguration update that changes service CIDR, there might be temporarily multiple DNS services active at the same time.

It also explicitly allocates IPs for the DNS services to counter the problem of API server using the old service CIDR until it has stored it in the storage at which point it will start using the new IP range.
Unfortunately I have yet to find an issue on kubernetes issue tracker (if any) regarding this problem but I was able to trace it down to API server service IP storage handling where it appears to have two loops where one is allocating the IPs and another is supposed to persist the new service IP range configuration to storage (etcd).
For explicit IP allocation, I opted to copy the kubernetes package - let me know if you have a better idea.


build.assets/makefiles/base/agent/planet-agent.service
```diff
@@ -25,11 +25,11 @@ ExecStart=/usr/bin/planet agent \
         --election-enabled=${PLANET_ELECTION_ENABLED} \
         --initial-cluster=${PLANET_INITIAL_CLUSTER} \
         --kube-addr=https://${KUBE_APISERVER}:6443 \
-        --cluster-dns=${KUBE_CLUSTER_DNS_IP} \
```
a-palchikov, line 28: This seems to be unused so I removed it.

tool/planet/args.go
```diff
@@ -0,0 +1,82 @@
...
+	"fmt"
+	"net"
+	"strings"
+
+	kv "github.com/gravitational/configure"
+	"github.com/gravitational/trace"
+	"gopkg.in/alecthomas/kingpin.v2"
+)
+
+func cidrFlag(s kingpin.Settings) *cidr {
```
a-palchikov, line 28: Opted to reimplement the CIDR flag from the [original](https://github.com/gravitational/configure/blob/master/cidr.go#L42) as I needed to have access to the underlying `net.IPNet`.

tool/planet/agent.go
```diff
@@ -477,6 +503,47 @@ func status(c statusConfig) (ok bool, err error) {
 	return ok, nil
 }
 
+func validateKubernetesService(ctx context.Context, serviceCIDR net.IPNet) error {
+	return utils.Retry(ctx, math.MaxInt64, 5*time.Second, func() error {
```
r0mant, line 507: > math.MaxInt64

What's this? I assume you were going for an endless retry loop, so why not do it properly?

tool/planet/agent.go
```diff
@@ -477,6 +503,47 @@ func status(c statusConfig) (ok bool, err error) {
...
+		client, err := cmd.GetKubeClientFromPath(constants.SchedulerConfigPath)
```
r0mant, line 508: Does this method run on regular nodes too? It is called from the leader callback which doesn't seem to have any filters to run only on masters. Asking b/c regular nodes don't have scheduler config.

tool/planet/agent.go
```diff
@@ -477,6 +503,47 @@ func status(c statusConfig) (ok bool, err error) {
...
+			return trace.Wrap(err)
+		}
+		services := client.CoreV1().Services(metav1.NamespaceDefault)
+		svc, err := services.Get(KubernetesServiceName, metav1.GetOptions{})
+		if err != nil {
+			return trace.Wrap(err)
+		}
+		if svc.Spec.ClusterIP == "" {
+			// not a clusterIP service?
+			log.Info("kubernetes service clusterIP is empty.")
```
r0mant, line 519: Nit: **K**ubernetes

tool/planet/agent.go
```diff
@@ -477,6 +503,47 @@ func status(c statusConfig) (ok bool, err error) {
...
+			return nil
+		}
+		clusterIP := net.ParseIP(svc.Spec.ClusterIP)
+		if !serviceCIDR.Contains(clusterIP) {
+			log.WithField("cluster-ip", svc.Spec.ClusterIP).
+				Warn("kubernetes service clusterIP is invalid, will recreate service.")
```
r0mant, line 525: Nit: **K**ubernetes

Also, the message is a bit misleading since it doesn't actually recreate the service itself.

tool/planet/agent.go
```diff
@@ -477,6 +503,47 @@ func status(c statusConfig) (ok bool, err error) {
...
+			return removeKubernetesService(services)
+		}
+		log.Info("kubernetes service clusterIP is valid.")
```
r0mant, line 528: Nit: **K**ubernetes

Also, add current CIDR range and cluster IP to the message?

tool/planet/agent.go
```diff
@@ -477,6 +503,47 @@ func status(c statusConfig) (ok bool, err error) {
...
+		return nil
+	})
+}
+
+// Kubernetes defaults to the first IP in service range for the kubernetes service
+// See https://github.com/kubernetes/kubernetes/blob/v1.15.12/pkg/master/services.go#L41
+//
+// This checks the existing kubernetes services against the active service IP range
```
r0mant, line 536: I think this comment might be misplaced.

tool/planet/agent.go
```diff
@@ -477,6 +503,47 @@ func status(c statusConfig) (ok bool, err error) {
...
+		if svc.Spec.ClusterIP == "" {
+			// not a clusterIP service?
+			log.Info("kubernetes service clusterIP is empty.")
+			return nil
+		}
+		clusterIP := net.ParseIP(svc.Spec.ClusterIP)
+		if !serviceCIDR.Contains(clusterIP) {
+			log.WithField("cluster-ip", svc.Spec.ClusterIP).
+				Warn("kubernetes service clusterIP is invalid, will recreate service.")
+			return removeKubernetesService(services)
```
r0mant, line 526: I'm curious, do you think this might need any checks that the service was actually recreated, after we nuked it?

tool/planet/agent.go
```diff
@@ -477,6 +503,47 @@ func status(c statusConfig) (ok bool, err error) {
...
+			return trace.Wrap(err)
+		}
+		if svc.Spec.ClusterIP == "" {
+			// not a clusterIP service?
+			log.Info("kubernetes service clusterIP is empty.")
+			return nil
+		}
+		clusterIP := net.ParseIP(svc.Spec.ClusterIP)
+		if !serviceCIDR.Contains(clusterIP) {
+			log.WithField("cluster-ip", svc.Spec.ClusterIP).
```
r0mant, line 524: Add CIDR to the fields as well?

tool/planet/dns.go
```diff
@@ -110,16 +107,48 @@ func updateEnvDNSAddresses(client *kubernetes.Clientset, role agent.Role) error
...
+// of the service, and configure kubelet with the correct DNS addresses before starting
+func createDNSService(ctx context.Context, name string, services corev1.ServiceInterface, ipalloc *ipallocator.Range) error {
+	ip, err := ipalloc.AllocateNext()
+	if err != nil && err != ipallocator.ErrFull {
+		return &backoff.PermanentError{Err: err}
+	}
+	return utils.RetryWithInterval(ctx, newUnlimitedExponentialBackoff(5*time.Second), func() error {
+		_, err = services.Create(newDNSService(name, ip.String()))
+		if err == nil || errors.IsAlreadyExists(err) {
+			return nil
```
r0mant, line 136: I would log success / already exists cases.

tool/planet/dns.go
```diff
@@ -110,16 +107,48 @@ func updateEnvDNSAddresses(client *kubernetes.Clientset, role agent.Role) error
...
+// The service object is managed by gravity, but we create a placeholder here, so that we can read the IP address
+// of the service, and configure kubelet with the correct DNS addresses before starting
+func createDNSService(ctx context.Context, name string, services corev1.ServiceInterface, ipalloc *ipallocator.Range) error {
+	ip, err := ipalloc.AllocateNext()
+	if err != nil && err != ipallocator.ErrFull {
+		return &backoff.PermanentError{Err: err}
+	}
+	return utils.RetryWithInterval(ctx, newUnlimitedExponentialBackoff(5*time.Second), func() error {
+		_, err = services.Create(newDNSService(name, ip.String()))
+		if err == nil || errors.IsAlreadyExists(err) {
```
r0mant, line 135: So if we're ignoring "already exists" error, what's taking care of recreating the DNS services from the new IP range? Is it done by Gravity?

tool/planet/dns.go
```diff
@@ -145,13 +174,104 @@ func createService(name string) error {
...
+	if err != nil {
+		panic(err.Error())
+	}
+	return selector
+}
+
+// isIPRangeMistmatchError detects whether the given error indicates that the suggested cluster IP
+// is from an unexpected service IP range. This can happen as long as the apiserver's repair
+// step did not commit the new service IP range configuration to the store (eg etcd)
+func isIPRangeMistmatchError(err error) bool {
```
r0mant, line 240: Nit: This method doesn't appear to be used anywhere.

tool/planet/dns.go
```diff
@@ -110,16 +107,48 @@ func updateEnvDNSAddresses(client *kubernetes.Clientset, role agent.Role) error
...
+	return nil
+}
 
-	service := &v1.Service{
+// createDNSService creates the kubernetes DNS service if it doesn't already exist.
+// The service object is managed by gravity, but we create a placeholder here, so that we can read the IP address
+// of the service, and configure kubelet with the correct DNS addresses before starting
+func createDNSService(ctx context.Context, name string, services corev1.ServiceInterface, ipalloc *ipallocator.Range) error {
+	ip, err := ipalloc.AllocateNext()
+	if err != nil && err != ipallocator.ErrFull {
```
r0mant, line 130: What does ErrFull mean in this case? And why is it not a terminal error here but a terminal error below (inside isIPAlreadyAllocatedError case)?

tool/planet/dns.go
a-palchikov: Yes, gravity manages the DNS services during the operation. The agent could in theory ensure that the DNS service with an IP address from the service cidr existed and forced it if not but there's also a race if an agent on a node where container configuration is yet to be updated restarts, it might invalidate the operation state so I chose to keep the modifications on one side.

tool/planet/dns.go
a-palchikov: Not really following but it did not need to be `PermanentError` here so I changed it to `trace.Wrap` instead.

tool/planet/agent.go
a-palchikov: Yes - I'll make it retry in the same loop to make sure the service has been created.

r0mant (approved):

PR #701, opened 2020-07-10
Kernel version check

bernardjkim:
### Description
Add a kernel version check with the minimum supported kernel version specified at `3.10.0-1127`.

### Linked tickets and PRs
* Updates https://github.com/gravitational/gravity/issues/1818
* Requires https://github.com/gravitational/satellite/pull/244


lib/monitoring/checkers.go
```diff
@@ -292,6 +292,14 @@ func addToMaster(node agent.Agent, config *Config, etcdConfig *monitoring.ETCDCo
 	}
 	node.AddChecker(systemPodsChecker)
 
+	// Minimum required Linux kernel version 3.10.0
+	node.AddChecker(monitoring.NewKernelChecker(monitoring.KernelVersion{
```
a-palchikov, line 296: Can you move this kernel version value to consts/vars section and reuse below?

a-palchikov (approved):

lib/monitoring/checkers.go
```diff
@@ -292,6 +292,14 @@ func addToMaster(node agent.Agent, config *Config, etcdConfig *monitoring.ETCDCo
 	}
 	node.AddChecker(systemPodsChecker)
 
+	// Minimum required Linux kernel version 3.10.0
```
r0mant, line 295: Nit: 3.10.0-1127. Same below.

r0mant (approved):

lib/constants/constants.go
```diff
@@ -98,4 +102,12 @@ const (
 var (
 	// GravityDataDir is the directory where gravity data is stored in planet
 	GravityDataDir = "/var/lib/gravity"
+
+	// MinKernelVersion is the minimum supported Linux kernel version -- 3.10.0-1127
```
knisbet, line 106: Nit: so this is maybe a bit miss-leading, and it's an artifact of the original bug report from @r0mant . We need to be careful here because all kernels aren't equal. So this check probably only works in this particular case, because I image redhat releases which is what this is meant to protect against are running the oldest kernels. But it's unsafe to see say an ubuntu kernel bug and simply increase this value.

So I would try and atleast document this in the comment to a further extent, to try and avoid someone accidentally using this and triggering some unintended side effects between the various distros and the in use kernels.

PR #714, opened 2020-07-31
[7.0.x] intermediate upgrades

a-palchikov:
This PR removes the obsolete commands to manage the etcd upgrade service.

### DIFF
--- a/lib/utils/utils.go
+++ b/lib/utils/utils.go
@@ -24,7 +24,6 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
-	"strings"
 	"syscall"
 	"unicode"
 
@@ -32,7 +31,6 @@ import (
 
 	"github.com/ghodss/yaml"
 	"github.com/gravitational/trace"
-	log "github.com/sirupsen/logrus"
 )
 
 // WriteHosts formats entries in hosts file format to writer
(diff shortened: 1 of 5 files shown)


tool/planet/constants.go
```diff
@@ -307,7 +307,8 @@ const (
 	DefaultPlanetReleaseFile = "/etc/planet-release"
 
 	// AssumeEtcdVersion is the etcd version we assume we're using if we're unable to locate the running version
-	AssumeEtcdVersion = "v2.3.8"
+	// This should be the least available version among the releases we can directly upgrade
+	AssumeEtcdVersion = "v3.3.22"
```
a-palchikov, line 311: Question for @knisbet - do you think it's safe to assume?

tool/planet/etcd.go
```diff
@@ -700,6 +702,10 @@ func getEtcdDataDir() (string, error) {
 	return getBaseEtcdDir(version), nil
 }
 
+func getBaseEtcdDir(version string) string {
```
a-palchikov, line 705: Another question for @knisbet. With the previous implementation, it is puzzling when is the `/ext/etcd` return is useful (e.g. when `version == AssumeEtcdVersion`) unless I'm missing anything:
```go
func getBaseEtcdDir(version string) string {
	p := DefaultEtcdStoreBase
	if version != AssumeEtcdVersion {
		p = filepath.Join(DefaultEtcdStoreBase, version)
	}
	return p
}
```

tool/planet/constants.go
r0mant: Are we going to support upgrades from any 5.5, or only from the latest? In the former case, earlier 5.5 versions included etcd v3.3.11 IIRC.

tool/planet/etcd.go
```diff
@@ -163,9 +154,34 @@ func etcdBackup(backupFile string, backupPrefix []string) (err error) {
...
 }
 
+func etcdMigrate(fromVersion, toVersion string) (err error) {
+	if fromVersion == "" {
+		fromVersion, _, err = readEtcdVersion(DefaultEtcdCurrentVersionFile)
+		if err != nil {
+			if !trace.IsNotFound(err) {
+				return trace.Wrap(err)
+			}
+			fromVersion = AssumeEtcdVersion
```
r0mant, line 164: Do we really need to assume anything anymore? I think all versions we're upgrading from (5.5) now should have the "etcd version" file?

tool/planet/etcd.go
```diff
@@ -308,41 +321,38 @@ func etcdMemberPeerList(ctx context.Context, client *etcdv3.Client, advertisePee
...
+	log.Info("Checking etcd service status.")
+	status, err := getServiceStatus(ETCDServiceName)
+	if err == nil {
+		log.Infof("%v service status: %v.", ETCDServiceName, status)
 		if status != "inactive" && status != "failed" {
-			return trace.BadParameter("%v must be disabled in order to run the upgrade. current status: %v", service, status)
+			return trace.BadParameter("%v must be disabled in order to run the upgrade. current status: %v", ETCDServiceName, status)
 		}
+	} else {
+		log.Warnf("Failed to query status of service %v. Continuing upgrade. Error: %v.", ETCDServiceName, err)
```
r0mant, line 355: Nit: WithError(err)

tool/planet/etcd.go
```diff
@@ -179,45 +195,42 @@ func etcdDisable(upgradeService, stopAPIServer bool) error {
...
+	ctx, cancel := context.WithTimeout(context.Background(), EtcdUpgradeTimeout)
+	defer cancel()
 	return trace.Wrap(enableService(ctx, ETCDUpgradeServiceName))
 }
 
+// etcdEnable enables a disabled etcd node
+func etcdEnable(joinToMaster string) error {
+	ctx, cancel := context.WithTimeout(context.Background(), 120*time.Minute)
+	defer cancel()
+	err := etcdInitJoin(ctx, joinToMaster)
```
r0mant, line 223: If we just use existing data dirs on all nodes, we won't need to join anything anywhere, no?

a-palchikov: Yeah, makes sense to remove.

tool/planet/etcd.go
a-palchikov: I would set a minimum version of 5.5 where this is a fact - removing the assume case will definitely simplify this.

lib/utils/utils.go
```diff
@@ -140,6 +142,135 @@ func ToJSON(data []byte) ([]byte, error) {
...
+	}
+	return nil
+}
+
+// FileOption defines a functional option to apply to specified path
+type FileOption func(path string) error
+
+// PermOption changes the file permissions on the specified file
+// to perm
+func PermOption(perm os.FileMode) FileOption {
```
r0mant, line 260: Nit: Not that it matters much, but I think using the word "Option" in the option is a bit redundant. I would just call them `WithFileMode()`, `WithOwner()`.

tool/planet/etcd.go
```diff
@@ -179,188 +184,68 @@ func etcdDisable(upgradeService, stopAPIServer bool) error {
...
-					"advertise_peer_url": advertisePeerURL,
-				}).Info("This node is already member of cluster")
-				isMember = true
-			}
-		}
-	}
-	return
+// etcdEnable enables a disabled etcd node
+func etcdEnable() error {
+	ctx, cancel := context.WithTimeout(context.Background(), 120*time.Minute)
```
r0mant, line 210: 2 hour timeout?

tool/planet/etcd.go
```diff
@@ -163,9 +152,25 @@ func etcdBackup(backupFile string, backupPrefix []string) (err error) {
...
+		"from": srcDir,
+		"to":   dstDir,
+	}).Info("Copy etcd data.")
+	return trace.Wrap(utils.CopyDirContents(srcDir, dstDir))
+}
+
 // etcdDisable disables etcd on this machine
 // Used during upgrades
-func etcdDisable(upgradeService, stopAPIServer bool) error {
+func etcdDisableUpgrade() error {
```
r0mant, line 167: Are these disable/enable upgrade service functions still used anywhere?

r0mant (approved):

tool/planet/etcd.go
a-palchikov: Copy/paste gone wrong.

tool/planet/etcd.go
a-palchikov: Not really - will remove incl. the commands using them.

PR #716, opened 2020-08-04
[7.0.x] Fix issue with waitForEtcdStopped function

UnderGreen:
Current implementation could hang forever because one could use
another etcd cluster inside docker containers(pods).

Backport of https://github.com/gravitational/planet/pull/715

### DIFF
--- a/tool/planet/etcd.go
+++ b/tool/planet/etcd.go
@@ -42,11 +42,11 @@ import (
 	etcdconf "github.com/gravitational/coordinate/config"
 	backup "github.com/gravitational/etcd-backup/lib/etcd"
 	"github.com/gravitational/trace"
-	ps "github.com/mitchellh/go-ps"
 	"github.com/sirupsen/logrus"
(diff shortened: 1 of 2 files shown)


tool/planet/etcd.go
```diff
@@ -773,14 +800,12 @@ loop:
 		case <-ticker.C:
 		}
 
-		procs, err := ps.Processes()
+		pid, err := systemctlGetPID(ctx, "etcd")
```
r0mant, line 803: If the goal is to be able to distinguish etcd running in planet vs etcd running inside a container, wouldn't checking `ppid == 1` to determine whether the process is the planet's etcd be more reliable? And simpler for that matter (just a slight change of existing condition).

My main concern with the implemented approach is that if we rely on some behavior based on source code findings ("from source code it looks like...", "from manual testing..." in the comments above), then this solution may be fragile and potentially break b/w systemd versions (esp. considering it's systemd...)

PR #727, opened 2020-08-19
(5.5) Update state directory checker to check uid/gid

r0mant:
This pull requests updates the existing storage checker that checks /var/lib/gravity directory to also check that its uid/gid are expected (match service user/group). Requires https://github.com/gravitational/satellite/pull/255. Refs https://github.com/gravitational/gravity/issues/2022.

### DIFF
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -23,132 +23,132 @@
 ignored = ["github.com/Sirupsen/logrus", "github.com/gravitational/planet/build/*"]
 
 [prune]
-  go-tests = true
-  unused-packages = true
+go-tests = true
+unused-packages = true
 
 [[override]]
-  name = "github.com/coreos/go-systemd"
-  version = "v16"
+name = "github.com/coreos/go-systemd"
+version = "v16"
 
 [[override]]
-  name = "bitbucket.org/ww/goautoneg"
-  # Bitbucket doesn't support git protocol, which causes dep to timeout trying to connect
-  # force to https url instead
-  source = "https://bitbucket.org/ww/goautoneg.git"
+name = "bitbucket.org/ww/goautoneg"
+# Bitbucket doesn't support git protocol, which causes dep to timeout trying to connect
+# force to https url instead
+source = "https://bitbucket.org/ww/goautoneg.git"
 
 [[override]]
-  name = "github.com/hashicorp/serf"
-  revision = "11bb88abf7b17f0b794b51416a9107d781e95f35"
+name = "github.com/hashicorp/serf"
+revision = "11bb88abf7b17f0b794b51416a9107d781e95f35"
 
 [[override]]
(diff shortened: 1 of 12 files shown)


Gopkg.toml
```diff
@@ -23,132 +23,133 @@
...
-  branch = "release-branch.go1.9"
+branch = "release-branch.go1.9"
+name = "golang.org/x/net"
 
 [[constraint]]
-  name = "github.com/gravitational/satellite"
-  version = "=5.5.22"
+name = "github.com/gravitational/satellite"
+#version = "=5.5.22"
+branch = "roman/5.5/dircheck"
```
r0mant, line 92: Sorry about all these whitespace changes, my editor decided to auto-format the file. The only real change here is this one to revendor satellite.

UnderGreen: One could add `?w=1` to URL to skip whitespace only changes. https://github.com/gravitational/planet/pull/727/files?w=1

r0mant: Huh, nice trick!

tool/planet/start.go
```diff
@@ -161,6 +161,8 @@ func start(config *Config, monitorc chan<- bool) (*runtimeContext, error) {
 		box.EnvPair{Name: EnvElectionEnabled, Val: strconv.FormatBool(config.ElectionEnabled)},
 		box.EnvPair{Name: EnvDNSHosts, Val: config.DNS.Hosts.String()},
 		box.EnvPair{Name: EnvDNSZones, Val: config.DNS.Zones.String()},
+		box.EnvPair{Name: EnvServiceUID, Val: config.ServiceUser.UID},
+		box.EnvPair{Name: EnvServiceGID, Val: config.ServiceUser.GID},
```
bernardjkim, line 165: When do we want to add EnvPairs and how are they being used?

bernardjkim (approved):

knisbet (approved):

tool/planet/start.go
r0mant: @bernardjkim They are basically used to propagate information from the planet container's start configuration into the container itself - these variables are written to `/etc/container-environment` file inside the planet container and as such are made available to everything running inside planet - in this particular case these variables will be read/used by the planet agent.

lib/monitoring/checkers.go
```diff
@@ -72,6 +73,10 @@ type Config struct {
 	LowWatermark uint
 	// HighWatermark is the disk usage critical limit percentage of monitored directories
 	HighWatermark uint
+	// ServiceUID is the UID of the service (planet) user.
+	ServiceUID string
```
a-palchikov, line 77: nit: it would be more practical to store these are final values that need no further processing (i.e. parsing) and do the validation as early as possible - e.g. when accepting values on command line. As far as I can tell these are always integers.

PR #729, opened 2020-08-20
[7.0.x] keep the package manifest in the resulting image

knisbet:
Copy the package manifest into `/etc/planet` to keep it consistent as part of the docker image.
Since this is the location where the `tele build` will pull the manifest from when translating the image into a gravity package if a custom system container has been configured.

Ports #717

(cherry picked from commit cb73d27fcf44af388aeea42f9da1e99ae5e95bd9)


r0mant (approved):

r0mant:
I think this is a duplicate of https://github.com/gravitational/planet/pull/723.

knisbet:
Closing in favor of #723 which does the same thing.

PR #731, opened 2020-08-25
don't configure coredns with localhost resolver when not specified in resolv.conf

knisbet:
Replaces https://github.com/gravitational/planet/pull/479
Updates https://github.com/gravitational/gravity/issues/613

Currently, our reader for resolv.conf emulates the underlying libc behaviour, where if there are no nameservers listed that 127.0.0.1 is used. This creates a problem for us, as we pass this onto the resolver we configure for planet, which tries to use 127.0.0.1 as an upstream resolver and starts to consume lots of CPU usage. Instead we indicate not to use defaults when trying to configure planet/coredns. This does imply if someone actually wants to use a localhost resolver within planet, then the resolver will need to be explicitly configured.

Testing this was a bit of a problem on GCE. It appears that if kubelet detects it's running on GCE, but isn't able to resolve the metadata server, that kubelet hangs in startup. Adding the name to coredns allowed startup to complete, so I don't think it's specifically related to this PR. Testing on AWS worked as intended.


lib/utils/dns.go
```diff
@@ -169,8 +170,10 @@ func DNSReadConfig(rdr io.Reader) (*DNSConfig, error) {
 		return nil, err
 	}
 
-	if len(conf.Servers) == 0 {
-		conf.Servers = append(conf.Servers, defaultNS...)
+	if !noDefaultNameservers {
```
r0mant, line 173: I would invert the condition (`addDefaultNameservers`) - double negative hurts the brain.

r0mant (commented):
Do we know what actually causes CoreDNS CPU usage to spike in this case?

Is it specifically non-existent resolver (which would be strange), or a some sort of loop (for which it should presumably have a detection mechanism)? I'm just a little wary of changing the behavior w/o understanding (at least on my part) what causes it.

Also with this change - how does a user actually use localhost resolver if they want to? You mentioned the should configure it explicitly - does that mean in the host's resolv.conf, aren't we filtering those out?

knisbet:
> Do we know what actually causes CoreDNS CPU usage to spike in this case?

Not precisely, some assumptions are involved.

> Is it specifically non-existent resolver (which would be strange), 

I believe this is more likely than other explanations. It does appear to be alot higher CPU usage than I would expect for a missing resolver, but does make some sense when considering that every query that the instance isn't authoritative for would be proxied upstream and waiting for a response possibly in it's own goroutine.

When I reproduced the problem I only deleted the upstream nameservers, which means I still had a search domain enabled. Basically every internal query was still leaking via the search domain, and now timing out when forwarded via CoreDNS. I don't know how efficient CoreDNS is in this regard, but i do find it plausible. 

We're also not using any of the possible gates within CoreDNS for various reasons:
- We don't health_check the customers DNS: The argument is basically that inevitably we'll run into a customers DNS server that doesn't offer the standard response, and not have a way to turn it off.
- Maximum concurrent queries to an upstream: This would cap the number of queries the server would allow outstanding to a particular upstream, and return an immediate SERVFAIL. It might prevent the CPU usage by processing less requests, but we still have an underlying architectural problem by configuring an upstream server that may not exist.
- Caching (specifically negative caching): Some requests could be served via cache instead of every request going to upstream. Originally when writing this configuration, this didn't really seem beneficial as the query load within planet is really an emulation of the way the host works (I expect my tools to work the same inside/outside of planet), and for docker to pull external images. With the search leaks we're seeing, there could be an argument to enable caching (and negative caching, but IDK if negative caching includes ServFail). Again this doesn't really avoid the architectural problem of configuring a forward server that may not exist.



>or a some sort of loop (for which it should presumably have a detection mechanism)? I'm just a little wary of changing the behavior w/o understanding (at least on my part) what causes it.

I don't see this being a loop as a likely explanation. Basically it comes down to, within planet we only bind to 127.0.0.2 and we forward to 127.0.0.1. The only way I could see this as plausible, is if CoreDNS is doing something clever with it's interpretation of 127.0.0.1 and deciding to forward that to itself despite not being bound to the address.

> Also with this change - how does a user actually use localhost resolver if they want to? You mentioned the should configure it explicitly - does that mean in the host's resolv.conf, aren't we filtering those out?

We only filter those out on the kube-dns configuration. CoreDNS within a pod binds to all interfaces/addresses, and as such when copying a localhost address from the hosts configuration to the pod, does create a loop. This filtering is arguably correct here, because it would be difficult for us to write this in a way that could access a localhost resolver on the host, without conflicting with programs on the host network namespace, or creating other knock on security problems.

With this change, to use a localhost resolver the user just needs to be explicit and have `nameserver 127.0.0.xx` in resolv.conf. 

If they have that in their config and don't have a localhost resolver they'll have all the same problems of course. But atleast it's explicitly in their control where they can fix the problem by changing the hosts configuration. Right now, you can't specify don't use an upstream resolver at all is the problem. 

We also have the special case for systemd-resolved, which binds to 127.0.0.53. For cluster-dns we read the legacy configuration and use the upstream resolvers from the legacy configuration directly. For planet, we can just forward to systemd-resolved, as it's in the same network namespace.

r0mant (commented):
Okay, it makes sense but I have a question about this:

> Testing this was a bit of a problem on GCE. It appears that if kubelet detects it's running on GCE, but isn't able to resolve the metadata server, that kubelet hangs in startup. Adding the name to coredns allowed startup to complete, so I don't think it's specifically related to this PR.

Do you mean adding the metadata server hostname to CoreDNS hostfile? Did you try install on GCE with and without your change to make sure there's no regression?

knisbet:
> Okay, it makes sense but I have a question about this:
> 
> > Testing this was a bit of a problem on GCE. It appears that if kubelet detects it's running on GCE, but isn't able to resolve the metadata server, that kubelet hangs in startup. Adding the name to coredns allowed startup to complete, so I don't think it's specifically related to this PR.
> 
> Do you mean adding the metadata server hostname to CoreDNS hostfile? Did you try install on GCE with and without your change to make sure there's no regression?

Correct, I added the metadata IP address to the hosts file coredns serves, and unblocked this startup problem. And yes, on my GCE cluster I did test the manual build with no modifications to the resolv.conf.

lib/utils/dns.go
```diff
@@ -89,7 +89,8 @@ func (d *DNSConfig) String() string {
 // See resolv.conf(5) on a Linux machine.
 // TODO(rsc): Supposed to call uname() and chop the beginning
 // of the host name to get the default search domain.
-func DNSReadConfig(rdr io.Reader) (*DNSConfig, error) {
+// noDefaultNameservers indicates we should not emulate libc behaviour and include a localhost resolver.
+func DNSReadConfig(rdr io.Reader, noDefaultNameservers bool) (*DNSConfig, error) {
```
a-palchikov, line 93: I'm a bit confused about the boolean parameter as well - do we need this or can we fall back to the suggested behavior by default - e.g. not include the defaults at all?

knisbet: This is the approach I attempted to take last year and the review process stalled there as well, so this is my attempt to explicitly indicate that the function violates normal behaviour so that it get's accepted.

https://github.com/gravitational/planet/pull/479

a-palchikov: The boolean is only used to pass true in the only use-case - why not just remove it and not include the default nameserver or am I missing anything? If this is a way to go I'd be fine with approving https://github.com/gravitational/planet/pull/479 instead.

a-palchikov (approved):

knisbet:
I can see it argued both ways.

This is in essence library code, which may be copied or re-used in some unpredictable way in the future, where the desired result may include emulating libc behaviour. So even though right now we only have a single use case where we don't want to emulate libc because it causes us problems, doesn't mean that's the only use case in some future of this code. 

This leaves several options:
- We can just not emulate libc, and try our best to document this and hope it doesn't bite us if something else needs the library code (the other PR that's been sitting in review for more than a year).
- We can be explicit, and indicate specifically via configuration that the behaviour we want does not represent libc's behaviour. If this was a from scratch implementation, the libc emulation could even just return not implemented errors, so if someone did try to use the code it's clear that the desired behaviour is not implemented (this PR).
- We can make this code not utility code, so it's tightly bound to the use case. This makes it less likely to be re-used, which is fine, because it's tightly coupled to the calling code which expects the behaviour. Doesn't preclude a mistake in the future where a refactor turns it back into library code which loses the context of this implementation being tied to the caller.

It seems like we don't really like any of the approaches, but we're otherwise just dancing in circles here. Do we want to be explicit about the behaviour, or do we want our library code tightly coupled to the expectations of the only known callers at present.

knisbet:
Temporarily setting health_check to a positive value, as it appears I was under a mistaken impression that setting a value of 0 would disable health checks. Instead it allows for atleast some of the runaway CPU usage.

PR #738, opened 2020-08-27
always use latest version of etcd when starting gateway

knisbet:
Updates https://github.com/gravitational/gravity/issues/2068

Fixes an oversight that crept in at some point, where the upgrade steps of etcd skip upgrading etcd when in gateway mode, but the gateway also doesn't just update when planet restart. Just make sure to use the latest version of etcd when in gateway mode (a worker) during startup.

### DIFF
--- a/tool/planet/etcd.go
+++ b/tool/planet/etcd.go
@@ -75,6 +75,18 @@ func etcdInit() error {
 			currentVersion = desiredVersion
 		}
 	}
+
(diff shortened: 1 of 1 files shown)


r0mant (approved):
lgtm if this works as expected

a-palchikov (approved):

knisbet:
I retested today upgrade 5.5.46 to latest, and it appears to work as intended.

PR #743, opened 2020-09-01
(6.1) Set systemd max files

r0mant:
https://github.com/gravitational/planet/pull/725 backport. Refs https://github.com/gravitational/gravity/issues/2090.

### DIFF
--- a/build.assets/makefiles/base/systemd/system.conf
+++ b/build.assets/makefiles/base/systemd/system.conf
@@ -1,0 +1,2 @@
+[Manager]
+DefaultLimitNOFILE=655350:655350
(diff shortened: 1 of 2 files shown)


knisbet (approved):

build.assets/makefiles/base/systemd/system.conf
```diff
@@ -0,0 +1,2 @@
+[Manager]
+DefaultLimitNOFILE=655350:655350
```
a-palchikov, line 2: Where does the colon come from - can you link to the docs on this syntax?

r0mant: From systemd [docs](https://www.freedesktop.org/software/systemd/man/systemd.exec.html):

> Resource limits may be specified in two formats: either as single value to set a specific soft and hard limit to the same value, or as colon-separated pair soft:hard to set both limits individually (e.g. "LimitAS=4G:16G").

Only in testing it appeared that specifying the single value didn't really override the soft limit so we set it explicitly.

PR #748, opened 2020-09-14
update iptables / satellite / flannel

knisbet:
1. Updates flannel to v0.10.2-gravitational
2. Updates satellite to latest master
3. Updates iptables to v1.8.5 (latest available)

Notes:
- I can't say I'm a huge fan of the way iptables get's built here. Going to a multi-stage build would require carefully tracking all the file copies from `make install` and could change between versions. The biggest help would be using buildkit, but that can't be implemented until our build servers support buildkit. 
- I also just picked the latest iptables, not sure if we should be considering an older version for some reason.

Requires Backport

Updates https://github.com/gravitational/gravity/issues/2122

### DIFF
--- a/build.assets/docker/base.dockerfile
+++ b/build.assets/docker/base.dockerfile
@@ -2,6 +2,7 @@ ARG PLANET_OS_IMAGE=planet/os
 FROM $PLANET_OS_IMAGE
 
 ARG SECCOMP_VER
+ARG IPTABLES_VER
 ARG PLANET_UID
 ARG PLANET_GID
 
@@ -55,6 +56,7 @@ RUN export DEBIAN_FRONTEND=noninteractive && set -ex && \
         && rm -rf /var/lib/apt/lists/*;
 
 # We need to use a newer version of iptables than debian has available
+# not ideal, but it's easier to run `make install` if we run this inline instead of a multi-stage build
 RUN export DEBIAN_FRONTEND=noninteractive && set -ex \
         && apt-get update \
         && apt-get install -q -y --allow-downgrades --no-install-recommends \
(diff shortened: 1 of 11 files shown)


build.assets/docker/base.dockerfile
```diff
@@ -55,6 +55,36 @@ RUN export DEBIAN_FRONTEND=noninteractive && set -ex && \
...
+        && apt-get update \
+        && apt-get install -q -y --allow-downgrades --no-install-recommends \
+        git \
+        autoconf \
+        libtool \
+        automake \
+        pkg-config \
+        libmnl-dev \
+        make \
+        && mkdir /tmp/iptables.build \
```
a-palchikov, line 70: nit: would it make sense to at least move this to a dedicated RUN step to keep it clearer?

a-palchikov (approved):

build.assets/docker/base.dockerfile
```diff
@@ -55,6 +55,36 @@ RUN export DEBIAN_FRONTEND=noninteractive && set -ex && \
...
+        && git clone git://git.netfilter.org/iptables.git --branch ${IPTABLES_VER} --single-branch /tmp/iptables.build \
```
r0mant, line 71: I'm curious why you chose to git clone it instead of just curling the tarball with the sources.

r0mant (approved):

build.assets/docker/base.dockerfile
knisbet: I went with this approach to try and limit the image size growth. While we may currently only take the resulting rootfs for packaging with gravity, I don't know if that will always be the case, and we do things like upload to quay.io for use in custom system containers. And I don't want to ship the additional tools in the resulting rootfs if we don't need them.

build.assets/docker/base.dockerfile
knisbet: No reason in particular, I don't see one approach vs the other having significant advantages. I suppose curl has a small advantage that we can set a bunch of retries for temporary failures. But at the same time I get corrupted curl downloads on my gce build nodes that I haven't been able to figure out. If we were using magnet, download the sources would be probably be preferable as that's much more setup for caching of downloaded artifacts, assuming the server is able to return an etag.

PR #752, opened 2020-09-17
Serf cluster reconciler

bernardjkim:
## Description
This PR adds a serf cluster reconcile loop to planet. This loop periodically reconciles the serf cluster in case a serf member is ever partitioned off the cluster. The implementation relies on the kubernetes api to determine if the serf cluster is missing any members.


a-palchikov:
This should also attempt to remove members that aren't part of the cluster as per [this ticket](https://github.com/gravitational/gravity/issues/943) since we don't have serf encryption and cannot prevent partially cleaned up nodes from rejoining if they still run the planet container.

tool/planet/agent.go
```diff
@@ -364,6 +364,7 @@ func runAgent(conf *agent.Config, monitoringConf *monitoring.Config, leaderConf
 		}
 	}
 
+	go startSerfReconciler(ctx, &conf.SerfConfig)
```
r0mant, line 367: I would modify the algorithm so it runs only on master nodes, otherwise we will be hammering API server in large clusters. We need to start keeping scaling in mind now.

tool/planet/membership.go
```diff
@@ -0,0 +1,157 @@
...
+	serf "github.com/hashicorp/serf/client"
+	log "github.com/sirupsen/logrus"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// startSerfReconciler creates a control loop that periodically attempts to
+// reconcile serf cluster.
+func startSerfReconciler(ctx context.Context, serfConfig *serf.Config) {
+	ticker := time.NewTicker(30 * time.Second)
```
r0mant, line 35: I think 30 seconds might be a bit too often for this type of check. I'd probably do it once every 5-10 minutes or so.

tool/planet/membership.go
```diff
@@ -0,0 +1,157 @@
...
+
+// getK8sPeers returns advertised IP addresses of all nodes in the kubernetes
+// cluster.
+func getK8sPeers() (peers []string, err error) {
+	client, err := monitoring.GetKubeClient()
+	if err != nil {
+		return peers, trace.Wrap(err, "failed to get kubernetes clientset")
+	}
+	nodes, err := client.CoreV1().Nodes().List(metav1.ListOptions{
+		LabelSelector: advertiseIPKey,
```
r0mant, line 83: What's this label selector for?

tool/planet/membership.go
```diff
@@ -0,0 +1,157 @@
...
+			log.WithField("k8s-peers", k8sPeers).
+				WithField("serf-peers", serfPeers).
+				Info("Reconciling serf cluster.")
+
+			numJoined, err := reconcileSerfCluster(serfConfig, k8sPeers)
+			if err != nil {
+				log.WithError(err).Warn("Failed to reconcile serf cluster.")
+				break
+			}
+			log.WithField("num-joined", numJoined).Info("Reconciled serf cluster.")
```
r0mant, line 66: I think logging the member that was missing and joined would be a bit more useful than just the number.

r0mant (commented):
> This should also attempt to remove members that aren't part of the cluster as per this ticket since we don't have serf encryption and cannot prevent partially cleaned up nodes from rejoining if they still run the planet container.

We need to be careful with this as this can potentially open a way to various race conditions where the agent e.g. may try to remove serf member while the new node is joining. I would keep it out of scope of what this PR tries to solve.

tool/planet/membership.go
bernardjkim: This is to ensure we are only selecting gravity nodes. Not sure if it is guaranteed that all nodes in the kubernetes cluster is part of the gravity cluster.

r0mant: Yes, we do not support adding non-Gravity-managed nodes to the cluster, so I think this filter is redundant.

r0mant (commented):
Have you tested how this behaves during shrink? To make sure it doesn't race with node removal and doesn't add the serf member back.

bernardjkim:
@r0mant 
As we discussed offline, the read order is now swapped and serf members is read before k8s nodes. 

We know that the k8s node is removed before the serf member is removed. So even though it is unlikely, if we encounter a race condition, at worst case, planet will just log that a serf member should no longer be a part of the cluster.

tool/planet/membership.go
```diff
@@ -0,0 +1,161 @@
...
+	k8sPeers, err := getK8sPeers(k8sClient)
+	if err != nil {
+		return trace.Wrap(err, "failed to get k8s peers")
+	}
+
+	if !shouldReconcileSerf(k8sPeers, serfPeers) {
+		return nil
+	}
+
+	if _, err := serfClient.Join(k8sPeers, false); err != nil {
```
r0mant, line 84: I'm just curious, why are we joining to all peers, instead of having `shouldReconcileSerf` return a list of missing peers and join just those?

Then you wouldn't also need the `getSerfPeers` call below as you'd just be able to log the serf cluster and the missing peers.

bernardjkim: Good point. No need to join all peers.

tool/planet/membership.go
```diff
@@ -0,0 +1,161 @@
...
+func getK8sPeers(client kubernetes.Interface) (peers []string, err error) {
+	nodes, err := client.CoreV1().Nodes().List(metav1.ListOptions{})
+	if err != nil {
+		return peers, trace.Wrap(err, "failed to list kubernetes nodes")
+	}
+
+	for _, node := range nodes.Items {
+		addr, exists := node.Labels[advertiseIPKey]
+		if !exists {
+			return peers, trace.NotFound("node %s does not have %s label", node.Name, advertiseIPKey)
```
r0mant, line 107: I would just log the warning and skip over the node if it doesn't have this label for some reason and return the rest. Also, how about failing over to `status.addresses` field in case this label doesn't exist? And skipping over the node if that's empty too.

r0mant (approved):

PR #769, opened 2020-10-15
[5.5] update etcd-backup to avoid short timer

knisbet:


a-palchikov (approved):

Gopkg.toml
```diff
@@ -75,7 +75,8 @@ branch = "master"
 name = "github.com/gravitational/configure"
 
 [[constraint]]
-branch = "master"
+#branch = "master"
```
r0mant, line 78: Are you intending to change this back to master before merging? If not, remove the comment.

r0mant (approved):

Gopkg.toml
knisbet: Yea it's sort of in a weird state, I think their were merges in coordinate, which never came back to planet, so I had to just statically assign it for now. I suppose for now it makes sense to kick the can down the road, but the next person to touch coordinate may have a bit of work to do.

PR #783, opened 2020-10-23
[7.0.x] fix rootfs executable lookup for SELinux

a-palchikov:
lib/box: fix label selection for executables given without absolute paths.

When 'gravity exec' executed a command without specifying the absolute path to said command, it picked the wrong label since it attempted to match each command directly inside the `rootfs` directory.
Instead, if the command is given without absolute path, match it inside the `rootfs` directory by using the `PATH` value.

Before the change:
```
$ gravity exec systemctl status kube-kubelet
/bin/systemctl: permission denied
```

After the change:
```
$ gravity exec systemctl status kube-kubelet
<output of systemctl status kube-kubelet>
```

### DIFF
--- a/lib/box/selinux.go
+++ b/lib/box/selinux.go
@@ -18,6 +18,7 @@ package box
 
 import (
 	"bytes"
+	"os"
 	"os/exec"
 	"path/filepath"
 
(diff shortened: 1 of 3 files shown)


lib/box/selinux.go
```diff
@@ -33,19 +34,25 @@ import (
 // if the domain cannot be determined.
 // Assumes SELinux support is on
 func getSELinuxProcLabel(rootfs, cmd string) (label string) {
+	logger := log.WithField("path", cmd)
```
r0mant, line 37: Nit: Should this field maybe be added to the logger after the "abspath" has been determined? Just so it contains the actual command path.

r0mant (approved):

PR #784, opened 2020-10-27
OpenEBS improvements

lenko-d:
This PR provides the following OpenEBS related enhancements:

* Iscsid to be setup in Planet in such a way that it only starts when OpenEBS enabled deployment is present. ISCSI related services in Planet are disabled(masked) if OpenEBS is not enabled in the application manifest .

Updates https://github.com/gravitational/gravity/issues/1005

## Testing done

Verified that the services get masked/unmasked when OpenEBS is disabled/enabled.


build.assets/docker/base.dockerfile
```diff
@@ -81,13 +81,21 @@ RUN export DEBIAN_FRONTEND=noninteractive && set -ex \
...
         libmnl-dev \
-        make \
-        && apt-get -y autoclean && apt-get -y clean && apt-get autoremove -y \
-        && rm -rf /var/lib/apt/lists/*;
+        make
+
+# Deploy Ubuntu .deb in order to get .socket activated iscsid. Debian's iscsid is not socket activated.
+# Socket activation is needed in order to avoid developing custom service start logic when OpenEBS app is deployed.
+RUN cd /var/lib/dpkg/
+RUN wget  http://mirrors.kernel.org/ubuntu/pool/main/o/open-iscsi/open-iscsi_2.0.874-7.1ubuntu6_amd64.deb
```
r0mant, line 89: Let's define the iscsi package version in the top Makefile and propagate it here like we do for other component versions.

Also, nit: spotted an extra space after "wget".

build.assets/docker/base.dockerfile
```diff
@@ -81,13 +81,21 @@ RUN export DEBIAN_FRONTEND=noninteractive && set -ex \
...
         pkg-config \
         libmnl-dev \
-        make \
-        && apt-get -y autoclean && apt-get -y clean && apt-get autoremove -y \
-        && rm -rf /var/lib/apt/lists/*;
+        make
+
+# Deploy Ubuntu .deb in order to get .socket activated iscsid. Debian's iscsid is not socket activated.
+# Socket activation is needed in order to avoid developing custom service start logic when OpenEBS app is deployed.
+RUN cd /var/lib/dpkg/
```
r0mant, line 88: Hm, I'm not sure this has any effect. If you want to run a command inside a specific directory, you can either use a single RUN clause like `RUN cd /var/lib/dpkg && wget ... && apt install ...` or use `WORKDIR` directive. 

In this case I don't think it really matters which directory you run this from, so I think you can just remove this and/or use absolute paths if you want to download it to a specific directory.

build.assets/docker/base.dockerfile
```diff
@@ -81,13 +81,21 @@ RUN export DEBIAN_FRONTEND=noninteractive && set -ex \
...
+RUN apt-get -y autoclean && apt-get -y clean && apt-get autoremove -y \
+            && rm -rf /var/lib/apt/lists/*;
 
 RUN set -ex && \
     groupadd --system --non-unique --gid ${PLANET_GID} planet && \
     useradd --system --non-unique --no-create-home -g ${PLANET_GID} -u ${PLANET_UID} planet && \
     groupadd --system docker && \
     usermod -a -G planet root && \
-    usermod -a -G docker planet;
+    usermod -a -G docker planet;
```
r0mant, line 101: Nit: Missing newline.

build.assets/docker/base.dockerfile
a-palchikov: My two cents - please prefix each RUN statement with `set -ex && ` so the commands are emphasized when the image is built as it's easier to look for build errors that way.
Also, each RUN creates an additional image layer - not sure it's beneficial to split these as they all deal with the same task - pulling and installing the package.

a-palchikov: And all hardcoded versions should ideally be pulled and configurable from the top-level Makefile and passed as ARGs here.

build.assets/docker/os.dockerfile
```diff
@@ -44,14 +44,6 @@ RUN set -ex && \
 	systemd-remount-fs.service dev-hugepages.mount sys-fs-fuse-connections.mount \
 	getty.target console-getty.service;
 
-# TODO(r0mant): Disable *iscsi* services cause they may be running on host
-#               In the future we will need to enable them conditionally to
-#               be able to support OpenEBS cStor engine out of the box
-RUN systemctl mask iscsi.service && \
```
a-palchikov, line 50: Can you clarify what would happen now with iscsi service on starting the container? Is it disabled/masked by default?

a-palchikov (approved):

build.assets/docker/os.dockerfile
lenko-d: By default the iscsid service will not be started. It will be inactive. It will be started on-demand by the iscsid.socket systemd unit.

```
 systemctl status iscsid.service
● iscsid.service - iSCSI initiator daemon (iscsid)
   Loaded: loaded (/lib/systemd/system/iscsid.service; disabled; vendor preset: enabled)
   Active: inactive (dead)
     Docs: man:iscsid(8)

```

```
systemctl status iscsid.socket
● iscsid.socket - Open-iSCSI iscsid Socket
   Loaded: loaded (/lib/systemd/system/iscsid.socket; enabled; vendor preset: enabled)
   Active: active (listening) since Thu 2020-10-29 14:19:09 UTC; 6min ago
     Docs: man:iscsid(8)
           man:iscsiadm(8)
   Listen: @ISCSIADM_ABSTRACT_NAMESPACE (Stream)
   CGroup: /system.slice/iscsid.socket

```


After I run the test command below it gets activated:

```
iscsiadm -m discovery -t st -p 10.05.02.10

systemctl status iscsid.service
● iscsid.service - iSCSI initiator daemon (iscsid)
   Loaded: loaded (/lib/systemd/system/iscsid.service; disabled; vendor preset: enabled)
   Active: active (running) since Thu 2020-10-29 14:29:41 UTC; 22s ago
     Docs: man:iscsid(8)
  Process: 29004 ExecStartPre=/lib/open-iscsi/startup-checks.sh (code=exited, status=0/SUCCESS)
  Process: 29015 ExecStart=/sbin/iscsid (code=exited, status=0/SUCCESS)
 Main PID: 29019 (iscsid)
    Tasks: 3 (limit: 4915)
   CGroup: /system.slice/iscsid.service
           ├─29018 /sbin/iscsid
           ├─29019 /sbin/iscsid
           └─29020 /sbin/iscsid


```

knisbet (approved):

r0mant (approved):

lib/systemd/service.go
```diff
@@ -0,0 +1,58 @@
...
+
+	"github.com/gravitational/planet/lib/constants"
+)
+
+// MaskService masks a service. Deletes the unit file first.
+// Returns an error if the unit file doesn't exist.
+func MaskService(unitFilePath string) error {
+	err := os.Remove(unitFilePath)
+	if err != nil {
+		return err
```
a-palchikov, line 31: `trace.Wrap(err)` here and below.

lib/systemd/service.go
```diff
@@ -0,0 +1,58 @@
...
+}
+
+// DropIn creates a systemd drop-in unit
+func DropIn(dropInFileName string, dropInDir string, dropInContents string) error {
+	err := os.MkdirAll(dropInDir, os.FileMode(constants.OwnerGroupRWXOtherRX))
+	if err != nil {
+		return err
+	}
+
+	dropInFullPath := dropInDir + dropInFileName
```
a-palchikov, line 52: It is better to use `filepath.Join` unless you can always be sure that `dropInDir` is slash-terminated.

tool/planet/start.go
```diff
@@ -525,6 +532,25 @@ func setupEtcd(config *Config) error {
 	return nil
 }
 
+// setupISCSI runs ISCSI related setup tasks.
+// If OpenEBS is disabled in the config then
+// we disable(mask) the ISCSI related services.
+func setupISCSI(config *Config) error {
+	if !config.OpenEBSEnabled {
```
a-palchikov, line 539: Looks like the function is a no-op if `config.OpenEBSEnabled == true` - why not move the condition check outside and pass it just the rootfs path?

a-palchikov (approved):

tool/planet/start.go
lenko-d: setupISCSI contains all steps that are needed in order to setup ISCSI. The check !config.OpenEBSEnabled is closely related to the setup process. As such it would be better to leave it as is. Let it be encapsulated as part of the setupISCSI collection of steps.

wadells:
Closing until such point as a maintainer has priority to pick this up.

PR #785, opened 2020-10-28
Add pod-subnet-size flag

bernardjkim:
## Description
This PR adds a `--pod-subnet-size` flag to determine the size of the subnet allocated to each host. The default size is /24, the max value is /28, and it cannot be larger than the pod CIDR range itself. The pod subnet size can also be specified and read from the `PLANET_POD_SUBNET_SIZE` environment variable.


tool/planet/cfg.go
```diff
@@ -154,12 +156,39 @@ func (cfg *Config) checkAndSetDefaults() (err error) {
...
 	}
 	return nil
 }
 
+// verifyPodSubnetSize verifies that the subnet size is not too small and verifies
+// that the subnet size is not larger than the network CIDR range.
+func verifyPodSubnetSize(subnetSize int, cidr kv.CIDR) error {
+	// Verify requested subnet size is not too small.
+	if subnetSize > 30 {
+		return trace.BadParameter("max pod subnet size value is /30")
```
r0mant, line 175: Nit: Include requested subnet size in the message.

tool/planet/cfg.go
```diff
@@ -154,12 +156,39 @@ func (cfg *Config) checkAndSetDefaults() (err error) {
...
+	}
+
+	// Verify subnet size is not larger than the network CIDR range.
+	_, ipv4Net, err := net.ParseCIDR(cidr.String())
+	if err != nil {
+		return trace.Wrap(err, "failed to parse cidr")
```
r0mant, line 181: Nit: Include CIDR value in the message.

tool/planet/cfg.go
```diff
@@ -154,12 +156,39 @@ func (cfg *Config) checkAndSetDefaults() (err error) {
...
+	}
+
+	prefixSize, _ := ipv4Net.Mask.Size()
+	if subnetSize < prefixSize {
+		return trace.BadParameter("subnet size cannot be larger than the network CIDR range")
```
r0mant, line 186: Nit: "subnet" -> "pod subnet", also include subnet size and requested range in the message.

tool/planet/cfg.go
```diff
@@ -154,12 +156,39 @@ func (cfg *Config) checkAndSetDefaults() (err error) {
...
 		cfg.VxlanPort = DefaultVxlanPort
 	}
 	return nil
 }
 
+// verifyPodSubnetSize verifies that the subnet size is not too small and verifies
+// that the subnet size is not larger than the network CIDR range.
+func verifyPodSubnetSize(subnetSize int, cidr kv.CIDR) error {
+	// Verify requested subnet size is not too small.
+	if subnetSize > 30 {
```
r0mant, line 174: /30 will give 4 pods per node max? I don't think even our bare base image install will fit. @knisbet What do you think would be a reasonable minimum? We could probably use whatever customer S requested as a reference, like I would probably consider /27 or at most (least?) /28 as the minimum reasonable size.

knisbet: /28 appears to be the minimum accepted by flannel: https://github.com/gravitational/flannel/blob/master/subnet/config.go#L70-L74

/28 does seem too small for a normal gravity install, but would probably be OK when disabling most the builtin apps. So I would probably start with a /28.

tool/planet/cfg.go
```diff
@@ -154,12 +156,42 @@ func (cfg *Config) checkAndSetDefaults() (err error) {
...
 	return nil
 }
 
+// verifyPodSubnetSize verifies that the subnet size is not too small and verifies
+// that the subnet size is not larger than the network CIDR range.
+func verifyPodSubnetSize(subnetSize int, cidr kv.CIDR) error {
+	// The minimum subnet size accepted by flannel is /28:
+	// https://github.com/gravitational/flannel/blob/master/subnet/config.go#L70-L74
+	if subnetSize > 28 {
+		return trace.BadParameter("pod subnet is too small. Minimum useful network prefix is /28: %d", subnetSize)
```
r0mant, line 176: Nit: It's a bit inconsistent that here the `subnetSize` is formatted into the message and below it's attached as a field. I would pick one way and stick to it.

r0mant (approved):

PR #788, opened 2020-11-05
[6.1.x] Add pod-subnet-size flag

bernardjkim:
## Description
This PR adds a `--pod-subnet-size` flag to determine the size of the subnet allocated to each host. The default size is /24, the max value is /28, and it cannot be larger than the pod CIDR range itself. The pod subnet size can also be specified and read from the `PLANET_POD_SUBNET_SIZE` environment variable.

## Linked tickets and PRs
* Ports https://github.com/gravitational/planet/pull/785


lenko-d (approved):

tool/planet/cfg.go
```diff
@@ -154,12 +156,44 @@ func (cfg *Config) checkAndSetDefaults() (err error) {
...
 		cfg.VxlanPort = DefaultVxlanPort
 	}
 	return nil
 }
 
+// verifyPodSubnetSize verifies that the subnet size is not too small and verifies
+// that the subnet size is not larger than the network CIDR range.
+func verifyPodSubnetSize(subnetSize int, cidr kv.CIDR) error {
+	// The minimum subnet size accepted by flannel is /28:
+	// https://github.com/gravitational/flannel/blob/master/subnet/config.go#L70-L74
```
a-palchikov, line 174: nit: use a specific commit if possible

tool/planet/cfg.go
```diff
@@ -79,6 +79,8 @@ type Config struct {
 	ServiceCIDR kv.CIDR
 	// PodCIDR defines the kubernetes Pod subnet CIDR
 	PodCIDR kv.CIDR
```
a-palchikov, line 81: nit: Somewhat unrelated but I faced the same issue with `kv.CIDR` and introduced a local flag that would expose the `net.IPNet` so you don't need to reparse it: https://github.com/gravitational/planet/blob/bernard/6.1.x/subnet-size/tool/planet/args.go#L28. This could save a couple of lines in `verifyPodSubnetSize`.

a-palchikov (approved):

PR #813, opened 2021-01-06
Add publishing to Drone CI

wadells:
## Summary
This changeset adds two further drone pipelines covering our two planet publishing scenarios.

Replacing https://jenkins.gravitational.io/job/Planet-dev-deploy/, the `trigger: event: custom` pipeline performs a dev deploy to s3 when manually invoked with a command like the following:

```
drone build create gravitational/planet --branch walt/drone-publish
```

**There is no way to trigger a dev deploy via the Drone UI**

Replacing https://jenkins.gravitational.io/job/Planet-Deploy-Artifacts/, the "trigger: event: tag" pipeline will publish to both s3 and quay every time a tag is pushed.

**Again, there is no way to trigger a full deploy via the Drone UI.**  It "just happens" when a tag is pushed.

## Testing Done

I pushed the `7.1.22-11709-test` tag to test the publishing job. Find the build logs and artifacts below:

https://drone.gravitational.io/gravitational/planet/7
https://quay.io/repository/gravitational/planet?tab=tags
https://s3.console.aws.amazon.com/s3/buckets/clientbuilds.gravitational.io?region=us-east-1&prefix=planet/7.1.22-11709-test/&showversions=false

I used the following drone command to test the dev publishing:

```
drone build create gravitational/planet --branch walt/drone-publish 
```

Resulting in:

https://drone.gravitational.io/gravitational/planet/10
https://s3.console.aws.amazon.com/s3/buckets/clientbuilds.gravitational.io?region=us-east-1&prefix=planet/7.1.22-11709-4-g591e4fd1/&showversions=false

## Notes
The AWS credentials are limited to s3 read write on the `clientbuilds.gravitational.io/planet` bucket/path.  They are managed in code, provisioned in https://github.com/gravitational/cloud-terraform/pull/109. The quay credentials are artisanally hand provisioned, and can be found under `gravitational+drone_planet_publish` here:

https://quay.io/organization/gravitational?tab=robots

## TODO
 - [x] Address PR feedback
 - [x] Write transition docs, send email to dev team
 - [x] Backport to 7.0
 - [x] Backport to 6.1
 - [x] Backport to 5.5
 - [x] Turn off Jenkins jobs
 - [x] Delete 7.1.22-11709-test  in quay
 - [x] Delete 7.1.22-11709-test  in github
 - [x] Delete 7.1.22-11709-test  in s3

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -118,8 +118,74 @@ services:
 volumes:
   - name: dockersock
     temp: {}
+
+---
+kind: pipeline
+type: kubernetes
+name: dev-publish
+
+trigger:
+  event:
+  - custom
+
+steps:
+  - name: fetch tags
+    image: docker:git
+    commands:
+      - git fetch --tags
+  - name: wait for docker
+    image: docker
+    commands:
+      - timeout 15s /bin/sh -c 'while [ ! -S /var/run/docker.sock ]; do sleep 1; done'
+      - docker version
+    volumes:
+      - name: dockersock
+        path: /var/run
+  - name: build buildbox
+    image: docker:git
+    commands:
+      - apk add --no-cache make
+      - make buildbox
+    volumes:
+      - name: dockersock
+        path: /var/run
+  - name: build tarball
+    image: docker:git
+    commands:
+      - apk add --no-cache make fakeroot
+      - make production
+    volumes:
+      - name: dockersock
+        path: /var/run
+  - name: publish to s3
+    image: docker:git
+    environment:
+      AWS_ACCESS_KEY_ID:
+        from_secret: AWS_ACCESS_KEY_ID
+      AWS_SECRET_ACCESS_KEY:
+        from_secret: AWS_SECRET_ACCESS_KEY
+      AWS_DEFAULT_REGION: us-east-1
+    commands:
+      - apk add --no-cache make aws-cli
+      - make dev-deploy
+    volumes:
+      - name: dockersock
+        path: /var/run
+
+services:
+  - name: run docker daemon
+    image: docker:dind
+    privileged: true
+    volumes:
+      - name: dockersock
+        path: /var/run
+
+volumes:
+  - name: dockersock
+    temp: {}
+
 ---
 kind: signature
-hmac: 16ff6331fc38270a5038e000f3687835b826ff6d2bb14d51c6702a6889b9342d
+hmac: 89b749595d4aac174eb23a4f8a4678b67241f2a05ada27ed6d02fea92a5acaae
 
 ...
--- a/.drone.yml
+++ b/.drone.yml
@@ -48,8 +48,78 @@ services:
 volumes:
   - name: dockersock
     temp: {}
+
+---
+kind: pipeline
+type: kubernetes
+name: publish
+
+trigger:
+  event:
+  - tag
+
+steps:
+  - name: fetch tags
+    image: docker:git
+    commands:
+      - git fetch --tags
+  - name: wait for docker
+    image: docker
+    commands:
+      - timeout 15s /bin/sh -c 'while [ ! -S /var/run/docker.sock ]; do sleep 1; done'
+      - docker version
+    volumes:
+      - name: dockersock
+        path: /var/run
+  - name: build buildbox
+    image: docker:git
+    commands:
+      - apk add --no-cache make
+      - make buildbox
+    volumes:
+      - name: dockersock
+        path: /var/run
+  - name: build tarball
+    image: docker:git
+    commands:
+      - apk add --no-cache make fakeroot
+      - make production
+    volumes:
+      - name: dockersock
+        path: /var/run
+  - name: publish
+    image: docker:git
+    environment:
+      REGISTRY_USERNAME:
+        from_secret: quay_username
+      REGISTRY_PASSWORD:
+        from_secret: quay_password
+      AWS_ACCESS_KEY_ID:
+        from_secret: AWS_ACCESS_KEY_ID
+      AWS_SECRET_ACCESS_KEY:
+        from_secret: AWS_SECRET_ACCESS_KEY
+      AWS_DEFAULT_REGION: us-east-1
+    commands:
+      - apk add --no-cache make aws-cli
+      - docker login -u="$REGISTRY_USERNAME" -p="$REGISTRY_PASSWORD" quay.io
+      - make deploy
+    volumes:
+      - name: dockersock
+        path: /var/run
+
+services:
+  - name: run docker daemon
+    image: docker:dind
+    privileged: true
+    volumes:
+      - name: dockersock
+        path: /var/run
+
+volumes:
+  - name: dockersock
+    temp: {}
 ---
 kind: signature
-hmac: b38e0d956bc9dc2abee7fb1418a05b7233ade75516da22cdfaceb01fb253ffce
+hmac: 16ff6331fc38270a5038e000f3687835b826ff6d2bb14d51c6702a6889b9342d
 
 ...


lenko-d (commented):
I noticed that there are 2 pipelines for 2 environments that are essentially the same source code. This introduces duplication. 

Is it going to be possible to parametrize the source code and make it to accept an ENV parameter? Maybe create a common pipeline that gets invoked by other pipelines for each ENV? Or some other Drone specific way to reduce duplication.

knisbet:
@wadells Feel free to add any requirements about integrations with a registry to https://github.com/gravitational/teleport/issues/5225

.drone.yml
```diff
@@ -48,8 +48,144 @@ services:
...
+        from_secret: quay_username
+      REGISTRY_PASSWORD:
+        from_secret: quay_password
+      AWS_ACCESS_KEY_ID:
+        from_secret: AWS_ACCESS_KEY_ID
+      AWS_SECRET_ACCESS_KEY:
+        from_secret: AWS_SECRET_ACCESS_KEY
+      AWS_DEFAULT_REGION: us-east-1
+    commands:
+      - apk add --no-cache make aws-cli
```
knisbet, line 103: I'm assuming apk commands aren't cached (and isn't just in this particular pipeline). Just a weird question, have you considered setting up our own mirror or building our own containers with all deps included? Just one of those things where we're block users pulling our files constantly, and might be hitting upstream more than we really need to?

Anyways, not important for this PR, I'm just curious if you're thinking about anything here, or maybe drone is already able to cache this which would be cool.

knisbet (approved):

bernardjkim (approved):

.drone.yml
wadells: > have you considered setting up our own mirror or building our own containers with all deps included?

Yes.  Particularly for the docker:git + make combo that is used in just about every step.

Upsides:

* Faster runtime (make is ~1s per step, so maybe 4-5s per pipeline). aws-cli is much heavier,  maybe 10-15s per download.
* Image caching across steps
* Reduce reliance on 3rd party infrastructure in our product builds, we'd depend on bits we host instead.


Downsides:
* We now have a 2nd code base feeding into these builds that needs to be maintained.  (See docker-debian) -- I've invested 1-2 weeks in maintaining/migrating that code over the past year.
* A new cron to fetch & publish these on a schedule. This is a semi-hidden moving part that would likely cause delayed failures if it tips over.
* Monitoring/alerting for the above if we want to be proactive about them working.

In short, I don't see the ROI at the moment.  This is work that definitely has value, but also increases the scope of the Jenkins migration if I choose to implement it as part fo my current work.  For now, I've filed https://github.com/gravitational/ops/issues/185 to track the request.

wadells:
> I noticed that there are 2 pipelines for 2 environments that are essentially the same source code. This introduces duplication.

The duplication is worse than this, as both of the publish pipelines are 70% duplicated from the pr pipeline.


> Is it going to be possible to parametrize the source code and make it to accept an ENV parameter? Maybe create a common pipeline that gets invoked by other pipelines for each ENV? Or some other Drone specific way to reduce duplication.

Yes. I mocked up a couple (untested) implementations:

https://github.com/gravitational/planet/blob/walt/drone-publish-env/.drone.yml#L109
https://github.com/gravitational/planet/blob/walt/drone-publish-when/.drone.yml#L93-L95

Take a look and see if either appeals to you.  There are a couple othe options that I didn't mock up because they require storing temporary artifacts in stable storage (s3).

*  using drone [pipeline dependencies](https://docs.drone.io/pipeline/configuration/#graph-execution) (data/artifacts aren't shared between pipelines, but they could be if pushed to s3 or any other bit bucket) to chain from validate to publish pipelines, using when conditionals to determine which links are chosen.
* using drone [promotions](https://docs.drone.io/promote/) to "upgrade" artifacts to a higher level -- more or less exactly intended for this dev/prod workflow.

I didn't chase either of these because they likely invlove storing development artifacts for a wide set of builds, and thus having some out of band clean up process or a much more rapidly growing set of builds to be stored indefinately.  Honestly, promotions seems like the most correct way to implement it, but I didn't want to scope creep to reworking our artifact storage.  I'm trying to stay focused on the path to turn jenkins off.

wadells:
Philosophically: I don't like putting complexity or significant logic into our CI configuration.  I'm of the opinion that CI configs should be as minimal as possible, and primarily wrap nuance implemented by the local build (Make).  This has the following advantages:


* Reproducing CI/publishing failures locally is a relatively simple task.  Just run the target that CI runs.
* Learning one tool (make, mage, et al) allows you to understand the important parts of the build for many projects.
* CI systems become relatively pluggable.  E.g. the migration from jenkins -> drone isn't nuanced.  If we put more logic into Jenkins/Drone, we make it harder to move to the next system.

With this in mind, I personally kind of like the "dumb as rocks" repetition of logic for each pipeline.  Yes, it is verbose and not at all DRY, but it is also accessible, explicit, and KISS.

knisbet:
> Philosophically: I don't like putting complexity or significant logic into our CI configuration. I'm of the opinion that CI configs should be as minimal as possible, and primarily wrap nuance implemented by the local build (Make). This has the following advantages:
> 
>     * Reproducing CI/publishing failures locally is a relatively simple task: Run the failing target CI from your workstation.
> 
>     * Learning one tool (make, mage, et al) allows you to understand the important parts of the build for many projects.
> 
>     * CI systems become relatively pluggable.  E.g. the migration from jenkins -> drone isn't nuanced.  If we put more logic into Jenkins/Drone, we make it harder to move to the next system.
> 
> 
> With this in mind, I personally like the "dumb as rocks" repetition of logic for each pipeline. Yes, it is verbose and not at all DRY, but it is also accessible, explicit, and KISS.

I agree, although maybe living with it is a different story, I'm currently a huge fan of this approach with my sort of fizzled push for mage. I sort of see CI/CD as a system for running actions based on git events, mapping some secrets, and being able to store the output, but otherwise really like the idea of the code itself in go. So in my ideal world, the drone pipeline is really just one thing, like `go run mage.go target` with required secrets and capturing/storing the output. 

But that's my 2 cents, I certainly haven't lived with this type of setup in the long run.

lenko-d:
>     * using drone [promotions](https://docs.drone.io/promote/) to "upgrade" a built to a higher level -- more or less exactly intended for this dev/prod workflow.

Promotions look very good to me. I would probably use that if it is easy to implement.

wadells:
> Promotions look very good to me. I would probably use that if it is easy to implement.

The ux problem I was working through with promotions is:

We need an initial build to promote.  The ways I've brainstormed for getting this are:

1) Opening a bogus PR to trigger the pr build config. This is undesirable as it is an abuse of PRs -- a mainstay of our development culture.  Out of the question IMO.
2) Trigger a build manually with `drone build create`.  This is the most targeted solution, but results in each publish requiring at least 2 invocations of the drone cli.  Not as nice as the one job jenkins workflow we have currently.  To make the experience nicer, this workflow could be wrapped in a make target, creating a bit of a CI ouroboros. Make calls drone calls make.
3) We build every commit pushed to the main repo.  This is nice verification, but results in a ton more builds than necessary -- especially with our culture of using a centralized repo for all sharing.  Planet has 81 branches, contrasted with 5 open PRs and 4 release branches.  Systemically, we only care about building commits tracked for merge.  The planet build is heavy enough that if we saw higher traffic this would be a notable waste of resources.  It'd definitely be out of the question for something as expensive as the gravity build.

After this initial build is created, the promote workflow has users run a 2nd `drone build promote` to dev or prod.  Again, here we either rebuild everything, or we need a non-drone cache of artifacts.  Rebuilding is wasteful of resources, and caching artifacts is non-trivial scope creep.

There is also something a bit wierd with tagging.  Production builds are expected to have a tag (though the automation doesn't require it) whereas dev builds may or may not.  The tag may be created after the initial `drone build create`, but before the `drone build promote`.  Using a promotion based workflow makes the "publish a release" step more complex.  It goes from
    push a tag
to
    push a tag
    drone build create
    drone build promote

We could have the deploy step create and push a tag, but my gut doesn't like giving drone push access to our repo -- I'd prefer to keep that limited to humans.

The other thing is, in terms of you initial duplicated config concern: promotion is largely a wash.  We'd still see duplicated pipelines, or branching using env/when.

In short, it is easy to implement a worse UX with a promotion workflow.  It is possible but hard to implement

Honestly, if we could drop the dev publish entirely (and have folks run that using their creds from their workstation), the CI problem would go away. We'd only have one publish, and it would be triggered on tag push. This raises the bar to entry though, and makes contributing to planet less friendly than it already is.  Probably not a good solution.

wadells:
Considering https://github.com/gravitational/planet/pull/813#issuecomment-756221619 and https://github.com/gravitational/planet/pull/813#issuecomment-755754037, I've examined several alternatives, and to me none are obviously better than what is implemented in this PR.  Each alternative has tradeoffs in complexity to implement or UX.

With that in mind, I'd like to merge the current implementation to start getting folks used to working with drone.  I propose we revisit this discussion in a couple weeks once I've had a chance to see what Gravity CI/release looks like and determine the appropriate patterns for that use case.  Overall, I'd like Planet CI workflows to be consistent with Gravity -- but I don't entirely know what Gravity will look like until I port it. I'm still building up experience with Drone.

What say you @lenko-d?

PR #821, opened 2021-01-30
Implement Kubernetes HA mode

bernardjkim:
## Description
This PR enables Kubernetes HA mode. 

A `--high-availability` boolean flag has been added to the planet `start` and `agent` commands. Default value will be read from container-environment variable `KUBE_HIGH_AVAILABILITY`.

If planet is running in HA mode, Kubernetes control plane components(`apiserver`, `controller-manager`, `scheduler`) will run on all master nodes vs running only on the elected leader.

### DIFF
--- a/tool/planet/agent.go
+++ b/tool/planet/agent.go
@@ -129,9 +129,9 @@ func startLeaderClient(config agentConfig, agent agent.Agent, errorC chan error)
 		return client, nil
 	}
 
-	// Watch for changes in the election mode.
-	// start/stop voter participation depending on the value of the election mode key.
-	client.AddWatchCallback(conf.ElectionKey, conf.Term, manageParticipation(client, conf, errorC))
+	// Watch for changes in the election key.
+	// Start/Stop voter participation when election is enabled/disabled.
+	client.AddWatchCallback(conf.ElectionKey, conf.Term, manageParticipation(client, config, errorC))
 
 	// Watch for changes to the leader key.
 	// If this agent becomes the leader, record a LeaderElected event to the local timeline.
@@ -139,7 +139,7 @@ func startLeaderClient(config agentConfig, agent agent.Agent, errorC chan error)
 
 	if !conf.HighAvailability {
 		// Watch for changes to the leader key.
-		// If this agent becomes the leader, start a number of additional services.
+		// Start/Stop control plane units when a new leader is elected.
 		client.AddWatchCallback(conf.LeaderKey, conf.Term/3, manageUnits(config))
 	}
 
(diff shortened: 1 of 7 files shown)


tool/planet/agent.go
```diff
@@ -143,85 +121,73 @@ func startLeaderClient(config agentConfig, agent agent.Agent, errorC chan error)
...
-	// watch the election mode status and start/stop participation
-	// depending on the value of the election mode key
-	client.AddWatchCallback(conf.ElectionKey, conf.Term, func(key, prevVal, newVal string) {
-		var err error
-		enabled, _ := strconv.ParseBool(newVal)
+
+	return client, nil
+}
+
+// manageParticipation starts voter participation when the election mode is
```
knisbet, line 168: nitpick: referring to this as election mode here and above reads a bit weird to me. I would consider rewording this a bit to just say whether elections are enabled/disabled.

knisbet (commented):
I suspect we should shutdown the control plane when the elections key for a particular node is disabled. See the note inline below about the particular issue this would help with.

tool/planet/agent.go
```diff
@@ -231,29 +197,72 @@ func startLeaderClient(config agentConfig, agent agent.Agent, errorC chan error)
 			cancelVoter()
 			cancelVoter = nil
 
+			// Control plane components run on all master nodes if HA mode is enabled.
+			if conf.HighAvailability {
+				return
```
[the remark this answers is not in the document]
bernardjkim: Updated in 495b0f2

While running with HA mode enabled , control plane units will be stopped when election is disabled, and units will start back up when election is enabled.

Behavior should be unchanged if running with HA mode disabled.

tool/planet/start.go
```diff
@@ -533,6 +534,12 @@ func addComponentOptions(config *Config) error {
 		config.Env.Append(EnvAPIServerOptions,
 			fmt.Sprintf("--service-node-port-range=%v", config.ServiceNodePortRange))
 	}
+	if config.HighAvailability {
+		config.Env.Append(EnvAPIServerOptions, fmt.Sprintf("--endpoint-reconciler-type=lease"))
```
a-palchikov, line 538:
```suggestion
		config.Env.Append(EnvAPIServerOptions, "--endpoint-reconciler-type=lease")
```
ditto below

tool/planet/agent.go
```diff
@@ -143,85 +121,75 @@ func startLeaderClient(config agentConfig, agent agent.Agent, errorC chan error)
...
 				return
 			}
 			// start election participation
+			var ctx context.Context
 			ctx, cancelVoter = context.WithCancel(context.TODO())
-			if err = client.AddVoter(ctx, conf.LeaderKey, conf.PublicIP, conf.Term); err != nil {
-				log.Errorf("failed to add voter for %v: %v", conf.PublicIP, trace.DebugReport(err))
+			err := client.AddVoter(ctx, config.leader.LeaderKey, config.leader.PublicIP, config.leader.Term)
+			if err != nil {
+				log.Errorf("failed to add voter for %v: %v", config.leader.PublicIP, trace.DebugReport(err))
```
a-palchikov, line 178:
```suggestion
				log.WithError(err),Errorf("Failed to add voter for %v.", config.leader.PublicIP)
```

a-palchikov, line 178: `cancelVoter()` as this is exiting after that

tool/planet/agent.go
```diff
@@ -143,85 +121,75 @@ func startLeaderClient(config agentConfig, agent agent.Agent, errorC chan error)
...
 				errorC <- err
 			}
+
+			// While running with HA enabled, start units when election is re-enabled
+			if config.leader.HighAvailability {
+				if err := startUnits(context.TODO()); err != nil {
```
a-palchikov, line 184: Use `ctx` from above

a-palchikov (approved):
I would also describe the mode of operation even briefly - e.g. that it will run local agent elections  and the coredns.hosts will contain just the elected node as apiserver.

a-palchikov:
One thing I did not notice is how the failed api server handled on the leader node (according to agent leader status)? Will it cause failover of the agent leader and updates to coredns.hosts file?

knisbet (approved):
lgtm with Dima's suggestion for removing the unneeded fmt.Sprintf's in a couple of locations.

bernardjkim:
@a-palchikov
If I understood your question correctly, there doesn't seem to be any failover in place if the apiserver fails. Would it make sense to watch the apiserver on the leader and run re-election in case the apiserver fails?

a-palchikov:
@bernardjkim Yes, I see this as a possible weak link. In case the apiserver fails on the leading agent node in HA mode, there will be no DNS update as long as the leader stays the same.

bernardjkim:
I'll go ahead and merge this PR for now. I'll work on implementing apiserver failover in a separate PR.

PR #824, opened 2021-02-26
Add cluster-dns script

bernardjkim:
## Description
This PR adds a `cluster-dns.sh` script that executes before the kubelet service is started. The script writes the cluster DNS servers into the kubelet config file. 

This is a necessary change before migrating to Kubernetes 1.19. Prior to 1.19, the cluster DNS servers are configured using the  `--cluster-dns` flag. This flag is now deprecated and the servers need to be configured through the kubelet config file.


build.assets/makefiles/master/k8s-master/cluster-dns.sh
```diff
@@ -0,0 +1,43 @@
...
+
+# If DNS_ADDRESSES is empty exit without any changes.
+if [ -z "$DNS_ADDRESSES" ]
+then
+    echo "DNS_ADDRESSES is not set"
+    exit
+fi
+
+# Remove existing clusterDNS values.
+perl -i -p0e 's/clusterDNS:(\n  - .*)*\n//g' $KUBELET_CONFIG
```
UnderGreen, line 34: I know it is up to you which tool to use here, but `sed`, in my opinion, is better known by anyone who will support the code and can provide the same result.

build.assets/makefiles/master/k8s-master/cluster-dns.sh
```diff
@@ -0,0 +1,43 @@
...
+
+# Write new clusterDNS values to kubelet config.
+CLUSTER_DNS="clusterDNS:"
+while IFS=',' read -ra ADDR; do
+    for i in "${ADDR[@]}"; do
+        CLUSTER_DNS+="\n  - $i"
```
UnderGreen, line 40: The same here - could be achieved by using sed or tee right here and don't call `echo -e` later.

build.assets/makefiles/master/k8s-master/cluster-dns.sh
bernardjkim: Thats fair, I've updated the script in 6e2b26f to use `sed` instead of `perl`. I'm not too familiar with `sed`, so let me know if the command can be simplified.

build.assets/makefiles/master/k8s-master/cluster-dns.sh
bernardjkim: What is the benefit of using `tee` over `echo`?

UnderGreen: I mean you can use sed/tee right here on line 40 without calling echo later

bernardjkim: I spent some time trying to figure out how to delete and append in the same command, but wasn't able to get it figured out. I don't want to spend too much time trying to learn how to use `sed` so I'll just leave it like this for now as long as it is producing the correct results.

build.assets/makefiles/master/k8s-master/cluster-dns.sh
```diff
@@ -0,0 +1,49 @@
...
+# This script writes the cluster DNS servers into the kubelet config file.
+#
+# More info on `--cluster-dns` can be found at: https://v1-19.docs.kubernetes.io/docs/reference/command-line-tools-reference/kubelet/
+
+# Environment file that contains the cluster DNS_ADDRESSES.
+DNS_ENV=/run/dns.env
+# Path to kubelet config file.
+KUBELET_CONFIG=/etc/kubernetes/kubelet.yaml
+
+# If /run/dns.env file does not exist exit without any changes.
```
knisbet, line 16: I believe it's possible to race during startup, so I think you want to exit with an error so that kubelet doesn't start until the file exists. 

Although I guess to be fair the file is still listed as an env file in the unit, so maybe this shouldn't really be possible unless something really funny occurs.

build.assets/makefiles/master/k8s-master/cluster-dns.sh
knisbet: I think that's fine if it's working.

knisbet (approved):

PR #833, opened 2021-03-25
Workaround PidsLimit config of systemd scope

knisbet:
Updates https://github.com/gravitational/gravity/issues/2444

Workaround for the -1 PidLimit not correctly passed to systemd by libcontainer. Newer runc updates does appear to fix this, however, updating runc in backports tends to be a bit risky so just set some large value we don't expect to see.

### DIFF
--- a/lib/box/srv.go
+++ b/lib/box/srv.go
@@ -388,7 +388,7 @@ func getLibcontainerConfig(containerID, rootfs string, cfg Config) (*configs.Con
 				AllowedDevices:   configs.DefaultAllowedDevices,
 				MemorySwappiness: nil, // nil means "machine-default" and that's what we need because we don't care
 				CpuShares:        2,   // set planet to minimum cpu shares relative to host services
-				PidsLimit:        -1,  // override systemd defaults and set planet scope to unlimited pids
+				PidsLimit:        2000000,  // override systemd defaults and set planet scope to unlimited pids
 			},
 		},
 		Devices:  append(configs.DefaultAutoCreatedDevices, append(loopDevices, disks...)...),


lib/box/srv.go
```diff
@@ -388,7 +388,7 @@ func getLibcontainerConfig(containerID, rootfs string, cfg Config) (*configs.Con
 				AllowedDevices:   configs.DefaultAllowedDevices,
 				MemorySwappiness: nil, // nil means "machine-default" and that's what we need because we don't care
 				CpuShares:        2,   // set planet to minimum cpu shares relative to host services
-				PidsLimit:        -1,  // override systemd defaults and set planet scope to unlimited pids
+				PidsLimit:        2000000,  // override systemd defaults and set planet scope to unlimited pids
```
a-palchikov, line 391: Is the value arbitrary? How is `-1` handled in libcontainer? If this is not using the same value - maybe use the same value?

a-palchikov (approved):

bernardjkim (approved):

lib/box/srv.go
knisbet: -1 is supposed to disable the limit, which I interpret as setting `max` in the cgroup as per: https://github.com/gravitational/planet/blob/master/vendor/github.com/opencontainers/runc/libcontainer/configs/cgroup_linux.go#L81-L82

However, when using the systemd backend the code only interprets a >0 value as per https://github.com/gravitational/planet/blob/master/vendor/github.com/opencontainers/runc/libcontainer/cgroups/systemd/apply_systemd.go#L234-L238 

The FS backend interprets this differently where it does properly support -1 as `max`: 
https://github.com/gravitational/planet/blob/master/vendor/github.com/opencontainers/runc/libcontainer/cgroups/fs/pids.go#L31-L42

So I'm interpreting this as a bug, where the FS backend using `-1` to set max/unlimitted, and a positive number to set that number. The systemd backend instead doesn't set the value, which allows the default to apply. The weird thing is systemd seems to still set this to unlimited, but only during a daemon-reload does it set the default value.

And then the latest version of libcontainer does seem to try and handle -1 but I'm still not convinced it's correct and last time I updated runc it was a giant pain: https://github.com/opencontainers/runc/blob/master/libcontainer/cgroups/systemd/v1.go#L87-L90

PR #836, opened 2021-04-15
Bump Kubernetes to 1.21.0

bernardjkim:
## Description
Bump Kubernetes from `v1.19.8` -> `v1.21.0`. Kubernetes changelogs [v1.20](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.20.md), [v1.21](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.21.md)

The `--kubelet-https` flag seems to have been deprecated a while ago, so the flag has been removed.

The ability to serve on an insecure port has been removed in v1.20, so the flag `--insecure-port` has been removed.
> The kube-apiserver ability to serve on an insecure port, deprecated since v1.10, has been removed. The insecure address flags --address and --insecure-bind-address have no effect in kube-apiserver and will be removed in v1.24. The insecure port flags --port and --insecure-port may only be set to 0 and will be removed in v1.24.

v1.20 introduced a new StorageVersionAPI alpha feature gate. Planet has all alpha features enabled by default, but this feature was causing kube-apiserver to crash with the current config. So I've disabled it for now.
> Add a StorageVersionAPI feature gate that makes API server update storageversions before serving certain write requests. This feature allows the storage migrator to manage storage migration for built-in resources. Enabling internal.apiserver.k8s.io/v1alpha1 API and APIServerIdentity feature gate are required to use this feature.

In v1.21, the `batch/v2alpha1` API group has been disabled. CronJobs have been promoted to `batch/v1` API group.
> The batch/v2alpha1 CronJob type definitions and clients are deprecated and removed. Promote CronJobs to batch/v1 

## Note
Kubernetes v1.20+ is needed to support our custom load balancer for Teleport Cloud https://github.com/gravitational/aws-global-accelerator-controller/issues/4

### DIFF
--- a/tool/planet/constants.go
+++ b/tool/planet/constants.go
@@ -434,7 +434,7 @@ const (
 	DefaultVxlanPort = 8472
 
 	// DefaultFeatureGates is the default set of component feature gates
-	DefaultFeatureGates = "AllAlpha=true,APIResponseCompression=false,BoundServiceAccountTokenVolume=false,CSIMigration=false,KubeletPodResources=false,IPv6DualStack=false,RemoveSelfLink=false,StorageVersionAPI=false"
+	DefaultFeatureGates = "AllAlpha=true,APIResponseCompression=false,StorageVersionAPI=false"
 
 	// DefaultServiceNodePortRange defines the default IP range for services with NodePort visibility
 	DefaultServiceNodePortRange = "30000-32767"
--- a/tool/planet/constants.go
+++ b/tool/planet/constants.go
@@ -434,7 +434,7 @@ const (
 	DefaultVxlanPort = 8472
 
 	// DefaultFeatureGates is the default set of component feature gates
-	DefaultFeatureGates = "AllAlpha=true,APIResponseCompression=false,BoundServiceAccountTokenVolume=false,CSIMigration=false,KubeletPodResources=false,IPv6DualStack=false,RemoveSelfLink=false"
+	DefaultFeatureGates = "AllAlpha=true,APIResponseCompression=false,BoundServiceAccountTokenVolume=false,CSIMigration=false,KubeletPodResources=false,IPv6DualStack=false,RemoveSelfLink=false,StorageVersionAPI=false"
 
 	// DefaultServiceNodePortRange defines the default IP range for services with NodePort visibility
 	DefaultServiceNodePortRange = "30000-32767"
--- a/Makefile
+++ b/Makefile
@@ -33,7 +33,7 @@ BUILDDIR ?= $(PWD)/build
 BUILDDIR := $(shell realpath $(BUILDDIR))
 OUTPUTDIR := $(BUILDDIR)/planet
 
-KUBE_VER ?= v1.19.8
+KUBE_VER ?= v1.21.0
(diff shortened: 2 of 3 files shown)


tool/planet/constants.go
```diff
@@ -434,7 +434,7 @@ const (
 	DefaultVxlanPort = 8472
 
 	// DefaultFeatureGates is the default set of component feature gates
-	DefaultFeatureGates = "AllAlpha=true,APIResponseCompression=false,BoundServiceAccountTokenVolume=false,CSIMigration=false,KubeletPodResources=false,IPv6DualStack=false,RemoveSelfLink=false"
+	DefaultFeatureGates = "AllAlpha=true,APIResponseCompression=false,BoundServiceAccountTokenVolume=false,CSIMigration=false,KubeletPodResources=false,IPv6DualStack=false,RemoveSelfLink=false,StorageVersionAPI=false"
```
helgi, line 437: Since you are touching these it might be a good time to clean these up unless there is a desire to keep them intentionally off still :) 

https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/

- `BoundServiceAccountTokenVolume` defaults to `true` in 1.21
- `CSIMigration` defaults to `true` as of 1.17
- `KubeletPodResources` has always been `true` actually, even since it was added as alpha in 1.13 (it's beta now) so perhaps a reason it was off and @knisbet or @r0mant remember why.
- `IPv6DualStack` defaults to `true` in 1.21 and went beta
- `RemoveSelfLink` defaults to`true` as of 1.20

bernardjkim: Thanks for pointing these out. I'll take a look and see what's safe to remove.

wadells:
I think before we can merge this we need to establish our plan/design for managing kubernetes versions and upgrade paths.  We have no viable upgrade path from 7.0 to 7.1 if gravity 7.0 is at kubernetes 1.17 and gravity 7.1 is at kubernetes 1.21.  IMO this deserves some design and documentation before we could merge this to master.

alternatively, we start 7.2.   But I would rather not try to port fixes across 5 active releases.

knisbet:
> Before we can merge this we need to establish our plan/design for managing kubernetes versions and upgrade paths. We have no viable upgrade path from 7.0 to 7.1 if gravity 7.0 is at kubernetes 1.17 and gravity 7.1 is at kubernetes 1.21. IMO this deserves some design and documentation.
> 
> Alternatively, we start 7.2. But I would rather not try to port fixes across 5 active releases.

I think we're just going to have to branch the current master into 7.1 and have 7.2 on the mast branch. I can't say I'm happy about it, but it's far easier than investing the engineering effort to try and see if we can be clever with the same branch building multiple versions of kubernetes at this point. Also we don't need to maintain any LTS guarentee's on any of these branches.

bernardjkim:
> Alternatively, we start 7.2. But I would rather not try to port fixes across 5 active releases.

It's might not be ideal, but I was thinking we just cut a 7.1 branch and start 7.2.

wadells:
> It's might not be ideal, but I was thinking we just cut a 7.1 branch and start 7.2.

Sounds good to me.

wadells (approved):

PR #838, opened 2021-05-10
Add build pipeline using a dockerfile

a-palchikov:
This PR adds a Docker-only build pipeline. It enables the builds on darwin.
It does not change the existing pipeline although the intention is to eventually replace it.


knisbet (commented):
Since this doesn't really interact with the internal build process, I'm not opposed.

Dockerfile
```diff
@@ -0,0 +1,463 @@
+# syntax = docker/dockerfile:1.2
+
+ARG KUBE_VER=v1.21.0
```
knisbet, line 3: One thing to watch out for is duplicating version variables might lead to some divergence. Might be worth considering picking a location and using it as a source of authority... otherwise the risk is to get out of sync.

knisbet (approved):

Dockerfile
a-palchikov: Yes, agree - I've been contemplating moving these to the makefile but left it as future improvement.

PR #839, opened 2021-05-15
Bump Kubernetes to 1.19.11

helgi:
https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.19.md#changelog-since-v11910

Libs have only been bumped up to 1.19.10, there is an RC for .11 but its been a few days without a release, .10 looks safe


helgi:
Drone doesn't handle it well when you switch the base branch after a PR is made. Recreating this PR.

PR #840, opened 2021-05-15
Bump Kubernetes to 1.19.11

helgi:
https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.19.md#changelog-since-v11910

Libs have only been bumped up to 1.19.10, there is an RC for .11 but its been a few days without a release, .10 looks safe


knisbet:
We're still vendoring deps, so I think the vendor dir needs to be updated.

helgi:
Vendored the packages

wadells (approved):
LGTM!

knisbet (approved):

PR #852, opened 2021-06-17
Kubernetes 1.19.12

helgi:
https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.19.md#v11912


wadells (approved):

wadells:
Did you revendor, I'm not seeing changes to `vendor/` akin to #853

helgi:
I did - 853 was using 1.19 libs which I moved to 1.21 so it vendored across
2 k8s minors.

Let me try vendoring again but don’t think it will produce much since
really was mostly a .11 -> .12

- Helgi

PR #867, opened 2021-10-15
Use certificates for kube-scheduler

UnderGreen:
After switching to using secure ports for controller-manager and scheduler the `/metrics` endpoint is not accessible by prometheus. More context can be found in that article why we do need authentication settings: https://octetz.com/docs/2018/2018-12-05-securing-controller-and-scheduler/.

Without those `--authentication-kubeconfig` and `--authorization-kubeconfig` RBAC configuration does not have any effect and when one tries to grab the metrics it is resulting in:
```
{
  "kind": "Status",
  "apiVersion": "v1",
  "metadata": {

  },
  "status": "Failure",
  "message": "forbidden: User \"system:anonymous\" cannot get path \"/metrics\"",
  "reason": "Forbidden",
  "details": {

  },
  "code": 403
}
```
Error from scheduler/CM:
```
r[3174]: W1015 14:06:31.230692    3174 authorization.go:156] No authorization-kubeconfig provided, so SubjectAccessReview of authorization tokens won't work
```


build.assets/makefiles/master/k8s-master/kube-controller-manager.service
```diff
@@ -14,12 +14,15 @@ ExecStart=/usr/bin/kube-controller-manager \
         --master=https://${KUBE_APISERVER}:6443 \
         --logtostderr=true \
         --kubeconfig=/etc/kubernetes/scheduler.kubeconfig \
+        --authentication-kubeconfig=/etc/kubernetes/scheduler.kubeconfig \
+        --authorization-kubeconfig=/etc/kubernetes/scheduler.kubeconfig \
         --profiling=false \
         --terminated-pod-gc-threshold=500 \
-        --port=0 \
```
bernardjkim, line 19: IIRC we set the --port=0 to disable the insecure port. Why does this need to be re-enabled?

build.assets/makefiles/master/k8s-master/kube-scheduler.service
```diff
@@ -9,7 +9,11 @@ ExecStartPre=/bin/systemctl is-active kube-apiserver.service
 ExecStart=/usr/bin/kube-scheduler \
         --master=https://${KUBE_APISERVER}:6443 \
         --kubeconfig=/etc/kubernetes/scheduler.kubeconfig \
+        --authentication-kubeconfig=/etc/kubernetes/scheduler.kubeconfig \
```
knisbet, line 12: I think you're going to need to add some context for these changes. Why the additional kubeconfig's to the base configuration.

build.assets/makefiles/master/k8s-master/kube-controller-manager.service
```diff
@@ -14,12 +14,15 @@ ExecStart=/usr/bin/kube-controller-manager \
         --master=https://${KUBE_APISERVER}:6443 \
         --logtostderr=true \
         --kubeconfig=/etc/kubernetes/scheduler.kubeconfig \
+        --authentication-kubeconfig=/etc/kubernetes/scheduler.kubeconfig \
```
knisbet, line 17: Should controller-manager really be authenticating as scheduler?

UnderGreen: And for the kubeconfig I will change it here for sure.

bernardjkim (approved):

wadells (approved):

PR #872, opened 2021-12-22
[8.0.x] Bump go to 1.17.5 (#871)

bernardjkim:
## Description
Bump go version to `1.17.5`.
Applied some changes that were required due to [module changes in Go 1.16](https://go.dev/blog/go116-module-changes).
- `go install` now requires a version suffix.
- Go defaults to `GO111MODULE=on`. We need to set `GO111MODULE=off` in places we want to continue using GOPATH mode.

## Linked tickets and PRs
- Ports https://github.com/gravitational/planet/pull/871
- Refs https://github.com/gravitational/gravity/issues/2693

### DIFF
--- a/versions.mk
+++ b/versions.mk
@@ -3,16 +3,16 @@ KUBE_VER ?= v1.19.15
 SECCOMP_VER ?= 2.3.1-2.1+deb9u1
 DOCKER_VER ?= 20.10.7
 # we currently use our own flannel fork: gravitational/flannel
-FLANNEL_VER ?= v0.10.5-gravitational
-HELM_VER ?= 2.16.12
-HELM3_VER ?= 3.3.4
-COREDNS_VER ?= 1.7.0
-NODE_PROBLEM_DETECTOR_VER ?= v0.6.4
-CNI_VER ?= 0.8.6
+FLANNEL_VER := v0.10.5-gravitational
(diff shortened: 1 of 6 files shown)


versions.mk
```diff
@@ -3,16 +3,16 @@ KUBE_VER ?= v1.19.15
...
-NODE_PROBLEM_DETECTOR_VER ?= v0.6.4
-CNI_VER ?= 0.8.6
+FLANNEL_VER := v0.10.5-gravitational
+HELM_VER := 2.16.12
+HELM3_VER := 3.3.4
+COREDNS_VER := 1.7.0
+NODE_PROBLEM_DETECTOR_VER := v0.6.4
+CNI_VER := 0.8.6
 SERF_VER ?= v0.8.5
-IPTABLES_VER ?= v1.8.5
```
wadells, line 13: Whats up with these changes?  Are they ported from somewhere or do we no longer need to be able to override these? I ask because I don't see them in https://github.com/gravitational/planet/pull/871.

wadells (approved):

versions.mk
bernardjkim: These changes were picked up when cherry picking commit to master. I'll revert these changes since it probably makes more sense to allow variable override.

PR #885, opened 2022-04-22
[7.0.x] Revert go upgrade

bernardjkim:
This reverts commit d96f1e9456b15da9f65d0f97e2dd1981f7fe2ed2. The Go upgrade resulted in volumes being unable to be mounted on Google Cloud https://github.com/gravitational/gravity/issues/2727. We'll revert the Go version to v1.12 until we can resolve the issue.


wadells (approved):
You may want to trim the extra (#XXX) declarations of of the commit message.  I find when they start to chain they get confusing.  Especially since another one will be added upon merge.

PR #886, opened 2022-04-26
[7.0.x] Update go-udev

bernardjkim:
## Description
Use `github.com/jochenvg/go-udev` instead of `github.com/gravitational/go-udev`. The go-udev package has been updated since we created a fork of the repo. 

Preemption implementation was changed in [Go 1.14](https://go.dev/doc/go1.14)

> A consequence of the implementation of preemption is that on Unix systems, including Linux and macOS systems, programs built with Go 1.14 will receive more signals than programs built with earlier releases. This means that programs that use packages like [syscall](https://go.dev/pkg/syscall/) or [golang.org/x/sys/unix](https://godoc.org/golang.org/x/sys/unix) will see more slow system calls fail with EINTR errors. Those programs will have to handle those errors in some way, most likely looping to try the system call again. For more information about this see [man 7 signal](http://man7.org/linux/man-pages/man7/signal.7.html) for Linux systems or similar documentation for other systems.

The latest commit contains https://github.com/jochenvg/go-udev/pull/4. 

> If the program that uses go-udev receives an interruption (SIGINT), go-udev stops monitoring and closes the channel. But this should only be the case when gracefully closed by done. Checking and continuing for EINTR would ensure that monitoring continues without stopping even if the source program receives any signal (usually when EINTR is seen).

I think Go1.14 is when the bug would have been introduced into Planet and the linked `go-udev` PR is the fix.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -554,6 +554,14 @@
   version = "v0.3.6"
 
 [[projects]]
+  branch = "master"
+  digest = "1:96e2e68cc63ed2010448ebbad799e16d25bcdc997a23dc7722bb5fb70d950be6"
+  name = "github.com/jkeiser/iter"
+  packages = ["."]
+  pruneopts = "UT"
+  revision = "c8aa0ae784d15ad3cb2a1f1579768e46f6e87fe5"
+
+[[projects]]
   digest = "1:e22af8c7518e1eab6f2eab2b7d7558927f816262586cd6ed9f349c97a6c285c4"
   name = "github.com/jmespath/go-jmespath"
   packages = ["."]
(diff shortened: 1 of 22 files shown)


wadells (approved):
LGTM.

It looks like @a-palchikov's changes were originally because https://github.com/jkeiser/iter wasn't licensed and thus a license contamination risk.  However, it has since been put under the MIT license: https://github.com/jkeiser/iter/commit/c8aa0ae784d15ad3cb2a1f1579768e46f6e87fe5, so we don't need https://github.com/gravitational/go-udev or its changes anymore.  At least, we won't once we get this change into maintained gravity branches.

PR #887, opened 2022-04-28
Update go-udev

bernardjkim:
## Description
Updates `go-udev` dependency. Fixes issue with udev monitor failing unexpectedly. Ports https://github.com/gravitational/planet/pull/886.


wadells (approved):
Thanks for the forward port!

PR #890, opened 2022-05-27
[7.0.x] Implement etcd-healthz check toggle

bernardjkim:
This PR adds a `etcd-healthz` check toggle using the `--etcd-healthz` flag. 
The changes also bumps `satellite` to modify the `etcd-healthz` checks to run in parallel. 

Addresses https://github.com/gravitational/gravity/issues/2740.
Requires https://github.com/gravitational/satellite/pull/305.

## TODO
- [x] Update upstream references / tags / versions after upstream PR merges (linked above)


wadells (approved):
LGTM.

Any thoughts about why go-udev is being removed?  Is that something we missed earlier?

Plese update the dependencies to the final version before merging.

bernardjkim:
> Any thoughts about why go-udev is being removed? Is that something we missed earlier?

We switched from `gravitational/go-udev` to `jochenvg/go-udev` in https://github.com/gravitational/planet/pull/886. The `gravitational/go-udev` dependencies should have been pruned in that PR, but I guess I didn't run the right `godep` commands to prune unused dependencies. They've been pruned here after revendoring `satellite`.

